PR #5, opened 2015-08-31
Ph-Tree implementation

tzaeschke:
Implementation of the PH-Tree. 
Properties: In-memory only, currently supports `DynamicIndex`,  `KNNIndex` and `RangeIndex`


elki/src/main/java/ch/ethz/globis/pht/PhDimFilter.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2011-2015 ETH Zurich. All Rights Reserved.
+ *
+ * This software is the proprietary information of ETH Zurich.
+ * Use is subject to license terms.
```
kno10: Here, the copying permission should be made explicit, as in the other ELKI files.
This is recommended by gnu.org ("copying permission statement"):
> To do so, attach the following notices to the program. It is safest to attach them to the start of each source file to most effectively state the exclusion of warranty; and each file should have at least the "copyright" line and a pointer to where the full notice is found.

"Proprietary information", "subject to license terms" does not state which license terms are applicable.

elki/src/main/java/ch/ethz/globis/pht/pre/ColumnType.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2011-2015 ETH Zurich. All Rights Reserved.
+ *
+ * This software is the proprietary information of ETH Zurich.
+ * Use is subject to license terms.
+ */
+package ch.ethz.globis.pht.pre;
+
+public abstract class ColumnType {
```
kno10: JavaDoc for this class would be good, in case someone wants to extend it to new column types. I know it isn't fun, and `depth` is not hard to understand.

elki/src/main/java/ch/ethz/globis/pht/pre/EmptyPP.java
```diff
@@ -0,0 +1,63 @@
...
+	}
+	
+	public EmptyPP(ColumnType[] types) {
+		this.types = Arrays.copyOf(types, types.length);
+	}
+	
+	@Override
+	public void pre(double[] raw, long[] pre) {
+		if (types != null) {
+			for (int d=0; d<raw.length; d++)
```
kno10: In ELKI code style, we do not save extra {} brackets; it saves only a few bytes/lines of source code, but is more likely to introduce errors (think Apples "goto fail;" SSL bug, which would have been prevented by using these brackets!)

elki/src/main/java/ch/ethz/globis/pht/pre/PreProcessorRangeD.java
```diff
@@ -0,0 +1,24 @@
...
+ * This software is the proprietary information of ETH Zurich.
+ * Use is subject to license terms.
+ */
+package ch.ethz.globis.pht.pre;
+
+public interface PreProcessorRangeD {
+	
+	/**
+	 * 
+	 * @param raw raw data (input)
```
kno10: @param raw should be @param raw1, @param raw2?

elki/src/main/java/ch/ethz/globis/pht/util/Bits.java
```diff
@@ -0,0 +1,19 @@
...
+
+
+/**
+ * Bit-stream manipulation functions.
+ * 
+ * @author ztilmann
+ *
+ */
+public class Bits extends BitsLong {
+
```
kno10: Empty?

elki/src/main/java/ch/ethz/globis/pht/util/PhIteratorBase.java
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright 2011-2014 ETH Zurich. All Rights Reserved.
+ *
+ * This software is the proprietary information of ETH Zurich.
+ * Use is subject to license terms.
+ */
+package ch.ethz.globis.pht.util;
+
+import java.util.Iterator;
+
+public interface PhIteratorBase<K, V, E> extends Iterator<V> {
```
kno10: This is the reason we don't use the Java `Iterator` API. Is the `extends Iterator<V>` sensible? This could also implement `Iterator<K>` and `Iterator<E>`, so it feels a bit dangerous to overload it this way.
Just a comment, no need to change this.

elki/src/main/java/ch/ethz/globis/pht/util/Refs.java
```diff
@@ -0,0 +1,182 @@
...
+//	private static int statACreate;
+//	private static int statATrim;
+	
+	public static final Object[] EMPTY_REF_ARRAY = {};
+    private static final ArrayPool POOL = 
+    		new ArrayPool(PhTreeHelper.ARRAY_POOLING_MAX_ARRAY_SIZE, 
+    				PhTreeHelper.ARRAY_POOLING_POOL_SIZE);
+
+    
+    private static class ArrayPool {
```
kno10: Array pooling is an interesting function. It may be a bit unfair though to only have this available in the PhTree, not in the other indexes, which could also benefit from object pooling and recycling. :-(

elki/src/main/java/de/lmu/ifi/dbs/elki/index/tree/spatial/ph/MinimalisticMemoryPHTree.java
```diff
@@ -0,0 +1,397 @@
...
+ * 
+ * @author Tilmann Zaeschke, Erich Schubert
+ * 
+ * @apiviz.has PHTreeKNNQuery
+ * @apiviz.has PHTreeRangeQuery
+ * 
+ * @param <O> Vector type
+ */
+@Reference(authors = "T. Zaeschke, C. Zimmerli, M.C. Norrie", title = "The PH-Tree -- A Space-Efficient Storage Structure and Multi-Dimensional Index", booktitle = "Proc. Intl. Conf. on Management of Data (SIGMOD'14), 2014", url = "http://dx.doi.org/10.1145/361002.361007")
+public class MinimalisticMemoryPHTree<O extends NumberVector> extends AbstractIndex<O>
```
kno10: In what sense is this tree "Minimalistic"? I'd rename this to `PHTree` or `MemoryPHTree`.
The reason our k-d-tree uses "Minimalistic" is because that version does not use Node objects, and is a much smaller implementation than used in popular kd-tree implementations; and we expect to eventually have other k-d-tree variants that e.g. choose axes by maximum extend instead (which requires keeping track of the splitting axis, and thus of "nodes").

elki/src/main/java/de/lmu/ifi/dbs/elki/index/tree/spatial/ph/MinimalisticMemoryPHTree.java
```diff
@@ -0,0 +1,397 @@
...
+    public PHTreeKNNQuery(DistanceQuery<O> distanceQuery, Norm<? super O> norm) {
+      super(distanceQuery);
+      this.norm = norm;
+    }
+
+    @Override
+    public KNNList getKNNForObject(O obj, int k) {
+      final KNNHeap knns = DBIDUtil.newHeap(k);
+      
+      for (double[] v: tree.nearestNeighbour(0, oToDouble(obj, new double[dims]))) {
```
kno10: shouldn't this be `nearestNeighbour(k, ...`? What about the distance function?

elki/src/main/java/de/lmu/ifi/dbs/elki/index/tree/spatial/ph/MinimalisticMemoryPHTree.java
```diff
@@ -0,0 +1,397 @@
...
+   * 
+   * For the PH-Tree, we perform a query on the bounding rectangle and then filter out all
+   * points that still violate the distance function.
+   * 
+   * TODO
+   * Unfortunately, this does not scale well with high dimensions because in high dimensions,
+   * cubes grow exponentially faster than spheres.
+   * As a solution we should pass in the distance function into the PH-query mechanism where
+   * it can be used to skip sub-trees that would fit into the rectangle but not into the given
+   * distance.
```
kno10: In many indexes, radius search can be implemented by doing a kNN search, and instead of stopping at k (+dulicates) neighbors stopping at the desired radius. This may be an option for high-dimensional data here, but often the distance-based pruning does not help that much either; the dimensionality problem exists in the distance function itself, too.

elki/src/main/java/de/lmu/ifi/dbs/elki/index/tree/spatial/ph/MinimalisticMemoryPHTree.java
```diff
@@ -0,0 +1,397 @@
...
+
+    DBIDIter iter = relation.getDBIDs().iter();
+
+    for(; iter.valid(); iter.advance()) {
+      O o = relation.get(iter);
+      double[] v = new double[dims];
+      for (int k = 0; k < dims; k++) {
+        v[k] = o.doubleValue(k);
+      }
+      DBID id = DBIDUtil.deref(iter);
```
kno10: I try to avoid materializing `DBID` objects because of the memory cost (16 bytes per ID, with 4 bytes of actual data - unfortunately this is Java and not C++). Algorithms shouldn't rely on the current `DBID` implementation, but I would be okay if an index only supports e.g. integer DBIDs, but at much lower memory cost and better performance. But that would of course be a larger change in your codebase.

kno10:
Since this is a quite substantial codebase (18k SLOC) with a different copyright owner (ETHZ). Furthermore, it does not require changes to the existing ELKI classes.
What do you think of keeping this tree in a separate module, similar to ELKI visualization, and the libsvm and lucene modules?

It would compile into its own artifact, with a dependency on ELKI core. This would make it easier to keep track of the copyright, and allow for independent updates. If you want, I can assist you with this. But essentially, it would just put all the files into a repository of its own, with a pom.xml that references ELKI (but of course, I would need to push ELKI 0.7.0 final to maven central to make it easy to build).
I have been considering similar restructuring to ELKI core, such as moving math functionality into a separate module, or indexes into an elki-indexing module. But some parts of ELKI are tightly coupled, e.g. the DeLiClu algorithm requires a particular index (a R*-tree variant), some algorithms come with their specialized distance functions, etc. so this is not easy to split into nice modules.

elki/src/main/java/ch/ethz/globis/pht/PhTree.java
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * k-dimensional index (quad-/oct-/n-tree).
+ * Supports key/value pairs.
+ *
+ *
+ * @author ztilmann (Tilmann Zaeschke)
+ *
+ */
+@Reference(authors = "T. Zaeschke, C. Zimmerli, M.C. Norrie", title = "The PH-Tree -- A Space-Efficient Storage Structure and Multi-Dimensional Index", booktitle = "Proc. Intl. Conf. on Management of Data (SIGMOD'14), 2014", url = "http://dx.doi.org/10.1145/361002.361007")
```
kno10: It may be okay to keep the @Reference annotation only on the "MinimalisticMemoryPHTree" class in the ELKI package, to avoid dependencies from ch.ethz to de.lmu.ifi.dbs.elki. Up to you to decide. Same in PhTree8.java. If we drop these annotations (it will still show up in the generated documentation; but only with one class name instead of three) there should be no dependency ch.ethz.* -> de.lmu.* anymore, and the MinimalisticMemoryPHTree class would be the sole interface between the two codebases.

elki/src/main/java/ch/ethz/globis/pht/PhDimFilter.java
tzaeschke: Done, see e20927a94b98d03bc6b2e8bb73c891652782e9d6
I also stated that "This file is part of ELKI". Let me know in case this is not good.

elki/src/main/java/ch/ethz/globis/pht/pre/ColumnType.java
tzaeschke: I removed the ColumnType, the way it was implemented it cause considerable loss of precision. The depth wasn't used at all. See 4f639edec2a990a361620540c750212ec557fd45

elki/src/main/java/ch/ethz/globis/pht/util/PhIteratorBase.java
tzaeschke: This is one of the things that comes from the full PhTree implementation. `Iterator` is not a perfect API, but it is well established and may help integrating the PH-Tree in other projects.

elki/src/main/java/ch/ethz/globis/pht/pre/EmptyPP.java
tzaeschke: I fully agree with using {} everywhere. This class was implemented by a student who wasn't aware of that. Let me know if you see any other missing {}.

elki/src/main/java/ch/ethz/globis/pht/pre/PreProcessorRangeD.java
tzaeschke: I went over the javadoc and fixed several issues with missing @param / @return, see d0d1273eb786706506e44bd3cc64fdb7fd95627c

elki/src/main/java/ch/ethz/globis/pht/util/Bits.java
tzaeschke: Yes. My internal working copy of the PhTree has different versions with different bit depths (long, int, byte). Using the above approach it is easy to switch between the representations when creating a copy of the tree.
I can remove it if there are considerable concerns, but it would make porting of future features somewhat more effort.

elki/src/main/java/ch/ethz/globis/pht/util/Refs.java
tzaeschke: Well:

1) Pooling can be disabled by setting `PhTreeHelper.ARRAY_POOLING_POOL_SIZE = 0`
2) Pooling wasn't used (or even implemented) in any paper I published so far, so no research is compromised.
3) It improves performance for the PhTree during insert/delete by 10-20%. This is because almost every insert/delete comes an array resize. This is necessary to reduce memory consumption of the many small nodes (most nodes have only two entries). Hence I doubt that other trees would gain as much as the PhTree from using pooling, i.e. the difference may not be measurable.

Anyway, I don't mind at all if we put this class in a more public place so it being used by other trees as well. 

As a side note, the tree I provide here has been significantly improved in many areas over the version that was published in my SIGMOD paper. None of the improvements have been published yet, I'm not sure I will find time anytime soon to publish them (they are not really worth an own publication anyway).

elki/src/main/java/de/lmu/ifi/dbs/elki/index/tree/spatial/ph/MinimalisticMemoryPHTree.java
tzaeschke: Agreed. See 3526628fa1f6f090aa7d22257e94e39a14f41205

elki/src/main/java/de/lmu/ifi/dbs/elki/index/tree/spatial/ph/MinimalisticMemoryPHTree.java
tzaeschke: Absolutely, nice catch. Fixed in 8d47d07eedaa2b8364ca593ea36395b4416b9dac

elki/src/main/java/de/lmu/ifi/dbs/elki/index/tree/spatial/ph/MinimalisticMemoryPHTree.java
tzaeschke: This could be implemented with the PhTree, in fact I have a distance function that checks for dominance of vectors (SkyLine queries). It currently requires Java 8 because it is implemented as a lambda. Once ELKI is on Java 8, this should be easy to implement.

But as I wrote in the comment, high dimensional cubes grow exponentially faster than spheres in high dimensions. Skipping sub-trees in the PH-tree will help a bit, but I don;t think it can fully solve the dimensionality problem.

elki/src/main/java/de/lmu/ifi/dbs/elki/index/tree/spatial/ph/MinimalisticMemoryPHTree.java
tzaeschke: If this is about the `sorted` field, I removed it, it was not necessary.

If it is about the loop: Probably I haven't really understood the design of ELKI yet. Could you show me how to access the data without materializing all the DBIDs?

Also, for the query results I need to return the database identifiers, I think. Since the PhTree only accepts Objects as values, I could Store an `Integer` object instead of the `DBID` objects, but I think that would not help much.

Do you have any suggestions? Did I understand you r comment correctly?

elki/src/main/java/ch/ethz/globis/pht/PhTree.java
tzaeschke: Done. See 341b450e471c986648e3b2b099807e4202aeb8d4

tzaeschke:
About putting the PhTree in its own module, I don't mind either way.
I could also wait until ELKI 0.7.0 is out or I could try to create a project anyway, maybe based on a SNAPSHOT release.

Where should the repo be located? I suppose it should be part of the ELKI repo?

elki/src/main/java/de/lmu/ifi/dbs/elki/index/tree/spatial/ph/MinimalisticMemoryPHTree.java
kno10: A `DBID` object is as expensive as an `Integer` object in Java; a substantial overhead. If the PhTree cannot store primitive `int` to conserve memory, it is okay as it is.

kno10:
We could make it entirely separate with a maven dependency on ELKI, or have it in the "addons/phtree" directory like other addons. We can set it up such that it compiles into its own jar, that you only need to add to the classpath to get the PhTree into ELKI (there is a service loader mechanism). The example script of the download package uses `-cp elki/*:depdenency/*`, so it will be possible to download the `phtree.jar`, drop it into either folder, and it should show up in minigui.

elki/src/main/java/ch/ethz/globis/pht/util/Bits.java
kno10: Ok, hotspot will deal with that with minimal overhead. No need to change.

kno10:
Restructured as an ELKI extension available at:

https://github.com/elki-project/extension-phtree/

PR #9, opened 2016-03-05
Added function to calculate cosine angle when one of the number vectors is sparse

vpranckaitis:


kno10:
Solved differently, in 4d5349b39c089d7cfdaa9fe230f4146beb1704d4
Also added a simple unit test and some code optimizations.

PR #11, opened 2016-04-15
Code quality fix - Declarations should use Java collection interfaces such as "List" rather than specific implementation.

faisal-hameed:
This pull request is focused on resolving occurrences of Sonar rule 
squid:S1319 - Declarations should use Java collection interfaces such as "List" rather than specific implementation.
You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S1319

Please let me know if you have any questions.

Faisal Hameed


kno10:
In many cases, one may want to use `ArrayList` because `LinkedList` is too expensive. Unfortunately, Java does not have an interface for lists-with-efficient-random-access.

faisal-hameed:
@kno10 I did't changed the implementation, it is still `ArrayList`. By using `List` as declaration gives developers liberty to change the implementation at any time without changing the actual code (method signature).

PR #14, opened 2016-04-15
Code quality fix - Local Variables should not be declared and then immediately returned or thrown.

faisal-hameed:
This pull request is focused on resolving occurrences of Sonar rule 
squid:S1488 - Local Variables should not be declared and then immediately returned or thrown. 
You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S1488

Please let me know if you have any questions.
 
Faisal Hameed

### DIFF
--- a/elki/src/main/java/de/lmu/ifi/dbs/elki/distance/distancefunction/strings/LevenshteinDistanceFunction.java
+++ b/elki/src/main/java/de/lmu/ifi/dbs/elki/distance/distancefunction/strings/LevenshteinDistanceFunction.java
@@ -104,8 +104,7 @@ public class LevenshteinDistanceFunction extends AbstractPrimitiveDistanceFuncti
         curr[j + 1] = Math.min(prev[j + 1] + 1, Math.min(curr[j] + 1, prev[j] + cost));
       }
     }
-    final int cost = curr[o2.length()];
-    return cost;
+    return curr[o2.length()];
   }
 
   @Override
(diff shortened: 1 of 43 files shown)


elki/src/main/java/de/lmu/ifi/dbs/elki/evaluation/clustering/PairCounting.java
```diff
@@ -138,8 +138,7 @@ protected PairCounting(ClusterContingencyTable table) {
    */
   public double fMeasure(double beta) {
     final double beta2 = beta * beta;
-    double fmeasure = ((1 + beta2) * pairconfuse[0]) / ((1 + beta2) * pairconfuse[0] + beta2 * pairconfuse[1] + pairconfuse[2]);
```
kno10: IMHO it makes sense to keep a named variable here, to emphasize the result of this equation is called the f-measure.

elki/src/main/java/de/lmu/ifi/dbs/elki/distance/distancefunction/strings/LevenshteinDistanceFunction.java
```diff
@@ -104,8 +104,7 @@ public static int levenshteinDistance(String o1, String o2) {
         curr[j + 1] = Math.min(prev[j + 1] + 1, Math.min(curr[j] + 1, prev[j] + cost));
       }
     }
-    final int cost = curr[o2.length()];
```
kno10: Here, too, I think a named variable `cost` does convey additional information for someone reading the code.

elki/src/main/java/de/lmu/ifi/dbs/elki/evaluation/clustering/PairCounting.java
faisal-hameed: @kno10 I agree with you but in 99% cases it is clear from the method name what is expected result e.g in this case method name is `fMeasure` which is expected value.

PR #18, opened 2016-04-19
Bugfix: fixed incorrrect typing in outlier histogram computation

patrickkostjens:
When using the automatic visualisation for LOF and other outlier algorithms, the following exception would be thrown:

    java.lang.ClassCastException: [D cannot be cast to de.lmu.ifi.dbs.elki.data.NumberVector
    	at de.lmu.ifi.dbs.elki.visualization.visualizers.visunproj.HistogramVisualization.makeVisualization(HistogramVisualization.java:96)
    	at de.lmu.ifi.dbs.elki.visualization.gui.overview.OverviewPlot.embedOrThumbnail(OverviewPlot.java:420)
    	at de.lmu.ifi.dbs.elki.visualization.gui.overview.OverviewPlot.reinitialize(OverviewPlot.java:322)
    	at de.lmu.ifi.dbs.elki.visualization.gui.overview.OverviewPlot.initialize(OverviewPlot.java:256)
    	at de.lmu.ifi.dbs.elki.visualization.gui.ResultWindow.<init>(ResultWindow.java:504)
    	at de.lmu.ifi.dbs.elki.result.AutomaticVisualization.processNewResult(AutomaticVisualization.java:112)
    	at de.lmu.ifi.dbs.elki.workflow.OutputStep.runResultHandlers(OutputStep.java:73)
    	at de.lmu.ifi.dbs.elki.KDDTask.run(KDDTask.java:121)
    	at de.lmu.ifi.dbs.elki.application.KDDCLIApplication.run(KDDCLIApplication.java:61)
    	at [...]

I tracked the problem to the creation of the histogram result which is expected to contain NumberVectors, but instead contained arrays of doubles. With this change, the visualisation works as expected.

PR #19, opened 2016-04-20
Bugfix: Histogram visualizer now uses double array instead of NumberVector

patrickkostjens:
This is a new fix for the same problem as in my previous pull request (#18). As per your suggestion there, I solved the problem in the visualization instead of the creation of the histogram.


kno10:
Thank you!
I've **merged** this as 42fbc8a9228614df659727e32e5b1d6f2bdf1f0f, but also removed the generic from the HistogramResult altogether in 792b8214f6223fec6b6b536b70ab9af16c6460e3. It was only used with `double[]` and probably cannot be used (meaningful) differently (`Double` would be quite expensive, and should be done differently). Maybe it will be further changed later.

PR #20, opened 2016-04-25
Bugfix: label rows are now also detected when preceded by comment rows

patrickkostjens:
If an input file starts with one or more comment lines, followed by a line containing labels, those labels are not parsed since originally labels would only be detected on line 1. Since `curvec` is only set when a vector row is parsed, it will always be `null` when the label row is parsed. This is therefore a more reliable way to detect label rows.

An alternative way to fix this would be to count non-comment line numbers in the `TokenizedReader`. However, this is a less elegant solution in my opinion while it also requires a closer coupling between `NumberVectorLabelParser` and `TokenizedReader`.


kno10:
**Merged** as bf9b83e5ecd11d370a1546138fdd8fe203ad18ed

PR #21, opened 2016-04-25
Bugfix: Clear labels after reading header row

patrickkostjens:
When the labels are not cleared after reading a header row, the first data row will have the titles in the header row as well as its own label set as the label. This causes the creation of an extra, non existing, cluster. Clearing the labels after creating the header row solves this problem.


kno10:
**Merged** as 7611152ecb6edc066995089ae2a1758ad67eaefd

PR #22, opened 2016-04-26
Partial code cleanup based on IDE warnings

patrickkostjens:
Based on warnings from my IDE, I did some code cleaning. The issues I fixed fall in the following categories:

- Use `System.arraycopy` instead of manual iteration.
- Use `Collections` methods instead of manual iterator or redundant array initialisation.
- Replace uses of deprecated methods.
- Replace non-looping `while` loops by `if`-statements.
- Remove redundant `if`-statements.
- Use `Objects`' `equals` method instead of string comparison using `==` or `!=`.
- Replace unnecessarily verbose `if`-statements.
- Remove redundant `return`-statements.

Additionally, I updated the copyright year for the files I modified.

### DIFF
--- a/addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/VisualizationTask.java
+++ b/addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/VisualizationTask.java
@@ -4,7 +4,7 @@ package de.lmu.ifi.dbs.elki.visualization;
  This file is part of ELKI:
  Environment for Developing KDD-Applications Supported by Index-Structures
 
- Copyright (C) 2015
+ Copyright (C) 2016
  Ludwig-Maximilians-Universität München
  Lehr- und Forschungseinheit für Datenbanksysteme
  ELKI Development Team
@@ -248,7 +248,7 @@ public class VisualizationTask implements VisualizationItem, Comparable<Visualiz
     // sort by name otherwise.
     String name1 = this.getMenuName();
     String name2 = other.getMenuName();
-    if(name1 != null && name2 != null && name1 != name2) {
+    if(name1 != null && name2 != null && !name1.equals(name2)) {
       return name1.compareTo(name2);
     }
     return 0;
--- a/addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/ResultWindow.java
+++ b/addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/ResultWindow.java
@@ -4,7 +4,7 @@ package de.lmu.ifi.dbs.elki.visualization.gui;
  This file is part of ELKI:
  Environment for Developing KDD-Applications Supported by Index-Structures
 
- Copyright (C) 2015
+ Copyright (C) 2016
  Ludwig-Maximilians-Universität München
  Lehr- und Forschungseinheit für Datenbanksysteme
  ELKI Development Team
@@ -465,10 +465,10 @@ public class ResultWindow extends JFrame implements ResultListener, Visualizatio
         if(e instanceof DetailViewSelectedEvent) {
           showSubplot((DetailViewSelectedEvent) e);
         }
-        if(OverviewPlot.OVERVIEW_REFRESHING == e.getActionCommand() && currentSubplot == null) {
+        if(OverviewPlot.OVERVIEW_REFRESHING.equals(e.getActionCommand()) && currentSubplot == null) {
           showPlot(null);
         }
-        if(OverviewPlot.OVERVIEW_REFRESHED == e.getActionCommand() && currentSubplot == null) {
+        if(OverviewPlot.OVERVIEW_REFRESHED.equals(e.getActionCommand()) && currentSubplot == null) {
           showOverview();
         }
       }
--- a/addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/SelectionTableWindow.java
+++ b/addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/SelectionTableWindow.java
@@ -3,7 +3,7 @@ package de.lmu.ifi.dbs.elki.visualization.gui;
 This file is part of ELKI:
 Environment for Developing KDD-Applications Supported by Index-Structures
 
-Copyright (C) 2015
+Copyright (C) 2016
 Ludwig-Maximilians-Universität München
 Lehr- und Forschungseinheit für Datenbanksysteme
 ELKI Development Team
@@ -290,10 +290,7 @@ public class SelectionTableWindow extends JFrame implements DataStoreListener, R
 
     @Override
     public boolean isCellEditable(int rowIndex, int columnIndex) {
-      if(columnIndex == 0) {
-        return false;
-      }
-      return true;
+      return columnIndex != 0;
     }
 
     @Override
--- a/addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/detail/DetailView.java
+++ b/addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/detail/DetailView.java
@@ -4,7 +4,7 @@ package de.lmu.ifi.dbs.elki.visualization.gui.detail;
  This file is part of ELKI:
  Environment for Developing KDD-Applications Supported by Index-Structures
 
- Copyright (C) 2015
+ Copyright (C) 2016
  Ludwig-Maximilians-Universität München
  Lehr- und Forschungseinheit für Datenbanksysteme
  ELKI Development Team
@@ -407,9 +407,7 @@ public class DetailView extends VisualizationPlot implements ResultListener, Vis
         vis.incrementalRedraw();
       }
     }
-    if(active || true) {
-      refresh();
-    }
+    refresh();
   }
 
   /**
--- a/addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/svg/SVGPath.java
+++ b/addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/svg/SVGPath.java
@@ -4,7 +4,7 @@ package de.lmu.ifi.dbs.elki.visualization.svg;
  This file is part of ELKI:
  Environment for Developing KDD-Applications Supported by Index-Structures
 
- Copyright (C) 2015
+ Copyright (C) 2016
  Ludwig-Maximilians-Universität München
  Lehr- und Forschungseinheit für Datenbanksysteme
  ELKI Development Team
@@ -29,6 +29,8 @@ import org.w3c.dom.Element;
 
 import de.lmu.ifi.dbs.elki.data.spatial.Polygon;
 import de.lmu.ifi.dbs.elki.utilities.datastructures.iterator.ArrayListIter;
+
+import java.util.Objects;
 
 /**
  * Element used for building an SVG path using a string buffer.
@@ -635,7 +637,7 @@ public class SVGPath {
    * @param ds coordinates.
    */
   private void append(String action, double... ds) {
-    if(lastaction != action) {
+    if(!Objects.equals(lastaction, action)) {
       buf.append(action);
       lastaction = action;
     }
@@ -651,7 +653,7 @@ public class SVGPath {
    * @return path object, for compact syntax.
    */
   public SVGPath close() {
-    if(lastaction != SVGConstants.PATH_CLOSE) {
+    if(!Objects.equals(lastaction, SVGConstants.PATH_CLOSE)) {
       buf.append(SVGConstants.PATH_CLOSE);
       lastaction = SVGConstants.PATH_CLOSE;
     }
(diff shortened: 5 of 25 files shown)


addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/VisualizationTask.java
```diff
@@ -248,7 +248,7 @@ public int compareTo(VisualizationTask other) {
     // sort by name otherwise.
     String name1 = this.getMenuName();
     String name2 = other.getMenuName();
-    if(name1 != null && name2 != null && name1 != name2) {
+    if(name1 != null && name2 != null && !name1.equals(name2)) {
       return name1.compareTo(name2);
```
kno10: Note that the actual comparison happens here. The identity check is a short-circuit, probably redundant.

addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/ResultWindow.java
```diff
@@ -465,10 +465,10 @@ public void actionPerformed(ActionEvent e) {
         if(e instanceof DetailViewSelectedEvent) {
           showSubplot((DetailViewSelectedEvent) e);
         }
-        if(OverviewPlot.OVERVIEW_REFRESHING == e.getActionCommand() && currentSubplot == null) {
+        if(OverviewPlot.OVERVIEW_REFRESHING.equals(e.getActionCommand()) && currentSubplot == null) {
           showPlot(null);
         }
-        if(OverviewPlot.OVERVIEW_REFRESHED == e.getActionCommand() && currentSubplot == null) {
+        if(OverviewPlot.OVERVIEW_REFRESHED.equals(e.getActionCommand()) && currentSubplot == null) {
```
kno10: Action commands are not free text strings, but constants. I think `==` should be used here.

addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/SelectionTableWindow.java
```diff
@@ -290,10 +290,7 @@ public String getColumnName(int column) {
 
     @Override
     public boolean isCellEditable(int rowIndex, int columnIndex) {
-      if(columnIndex == 0) {
-        return false;
-      }
-      return true;
+      return columnIndex != 0;
```
kno10: I might have a separate pull request open for such optimizations already that might cover this.

addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/detail/DetailView.java
```diff
@@ -407,9 +407,7 @@ protected void redraw() {
         vis.incrementalRedraw();
       }
     }
-    if(active || true) {
-      refresh();
-    }
+    refresh();
```
kno10: I'd prefer to keep this as is, and mark it with a FIXME - clearly, the `refresh()` was intended to be called only if `active` was set, but was force-enabled because of some synchronization issue probably - that may no longer persist.

addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/svg/SVGPath.java
```diff
@@ -651,7 +653,7 @@ private void append(String action, double... ds) {
    * @return path object, for compact syntax.
    */
   public SVGPath close() {
-    if(lastaction != SVGConstants.PATH_CLOSE) {
+    if(!Objects.equals(lastaction, SVGConstants.PATH_CLOSE)) {
```
kno10: Here, too, the actions are constants that can be safely compared with identity; which is more efficient.

addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/VisualizationTask.java
patrickkostjens: I missed that indeed, but removing the check I changed should be the solution then I think. That would make it more clear in my opinion.

addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/visualizers/scatterplot/TooltipStringVisualization.java
```diff
@@ -172,7 +172,7 @@ public Instance(VisualizationTask task, VisualizationPlot plot, double width, do
     protected Element makeTooltip(DBIDRef id, double x, double y, double dotsize) {
       final Object data = result.get(id);
       String label = (data == null) ? "null" : data.toString();
-      label = (label == "" || label == null) ? "null" : label;
+      label = (label == null || label.equals("")) ? "null" : label;
```
kno10: This should probably be merged!

addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/ResultWindow.java
patrickkostjens: In that case, it makes sense to leave it this way.

addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/SelectionTableWindow.java
patrickkostjens: I indeed did not check that. If that's the case, feel free to ignore this change.

elki-core-util/src/main/java/de/lmu/ifi/dbs/elki/utilities/optionhandling/parameterization/ChainedParameterization.java
```diff
@@ -58,9 +59,7 @@
    * @param ps Parameterizations
    */
   public ChainedParameterization(Parameterization... ps) {
-    for(Parameterization p : ps) {
-      chain.add(p);
-    }
+    Collections.addAll(chain, ps);
```
kno10: I don't like this extra method invocation with 0 benefits.

addons/batikvis/src/main/java/de/lmu/ifi/dbs/elki/visualization/gui/detail/DetailView.java
patrickkostjens: A FIXME or TODO would indeed be less confusing to show the intention of the `|| true`.

elki-docutil/src/main/java/de/lmu/ifi/dbs/elki/application/internal/DocumentReferences.java
```diff
@@ -333,7 +326,7 @@ public int compare(Pair<Reference, TreeSet<Object>> p1, Pair<Reference, TreeSet<
        * @return Order
        */
       private int compareNull(String s1, String s2) {
-        return (s1 == s2) ? 0 //
```
kno10: NO. The comparison happens below, the identity is only an optimization for the common case of `s1==s2`.

elki-gui-minigui/src/main/java/de/lmu/ifi/dbs/elki/gui/multistep/panels/ParameterTabPanel.java
```diff
@@ -313,10 +313,7 @@ protected synchronized void execute() {
    */
   public boolean canRun() {
     Status status = getStatus();
-    if (Status.STATUS_READY.equals(status) || Status.STATUS_COMPLETE.equals(status)) {
-      return true;
-    }
-    return false;
+    return Status.STATUS_READY.equals(status) || Status.STATUS_COMPLETE.equals(status);
```
kno10: This one is worth merging.

elki/src/main/java/de/lmu/ifi/dbs/elki/algorithm/clustering/kmeans/initialization/PredefinedInitialMeans.java
```diff
@@ -94,9 +94,7 @@ public void setInitialClusters(List<? extends Cluster<? extends MeanModel>> init
    */
   public void setInitialMeans(double[][] initialMeans) {
     double[][] vecs = new double[initialMeans.length][];
-    for(int i = 0; i < initialMeans.length; ++i) {
-      vecs[i] = initialMeans[i]; // TODO: clone?
-    }
+    System.arraycopy(initialMeans, 0, vecs, 0, initialMeans.length);
```
kno10: This change loses the TODO, which is not resolved.

elki/src/main/java/de/lmu/ifi/dbs/elki/data/BitVector.java
```diff
@@ -99,7 +99,7 @@ public boolean booleanValue(int dimension) {
   @Override
   @Deprecated
   public Bit getValue(int dimension) {
-    return new Bit(booleanValue(dimension));
+    return Bit.valueOf(booleanValue(dimension));
```
kno10: This should be merged.

elki/src/main/java/de/lmu/ifi/dbs/elki/datasource/parser/ClusteringVectorParser.java
```diff
@@ -152,7 +152,7 @@ public Event nextEvent() {
       return ret;
     }
     try {
-      while(reader.nextLineExceptComments()) {
+      if(reader.nextLineExceptComments()) {
```
kno10: No. May contain more than one result.

elki/src/main/java/de/lmu/ifi/dbs/elki/datasource/parser/SimpleTransactionParser.java
```diff
@@ -118,7 +118,7 @@ public Event nextEvent() {
       return ret;
     }
     try {
-      while(reader.nextLineExceptComments()) {
+      if(reader.nextLineExceptComments()) {
```
kno10: NO!

kno10:
Most of these changes are not good, and some may even introduce bugs, sorry.

Sometimes, the automatic recommendations of IDEs are not good.

elki-core-util/src/main/java/de/lmu/ifi/dbs/elki/utilities/optionhandling/parameterization/ChainedParameterization.java
patrickkostjens: I would personally prefer the Collections way because it shows the intention more clearly in my opinion, but the `for`-loop is already pretty clear in this case as well, so I guess the difference is small.

elki-docutil/src/main/java/de/lmu/ifi/dbs/elki/application/internal/DocumentReferences.java
patrickkostjens: You are right indeed.

elki/src/main/java/de/lmu/ifi/dbs/elki/algorithm/clustering/kmeans/initialization/PredefinedInitialMeans.java
patrickkostjens: I noticed the TODO, but I am not sure I understand what you still would want to accomplish with it. Or do your prefer a clone over an array copy?

elki/src/main/java/de/lmu/ifi/dbs/elki/datasource/parser/ClusteringVectorParser.java
patrickkostjens: I know that it may contain more than one result, but there is no way to get out of the `while`-loop without either returning a value or throwing an exception. Therefore, a next iteration of the `while`-loop will never be executed. I think the following lines are read by calling this method multiple times. 

As it currently stands, the `while` part is confusing in my opinion as it does not actually loop.

elki/src/main/java/de/lmu/ifi/dbs/elki/datasource/parser/SimpleTransactionParser.java
patrickkostjens: Same as above.

patrickkostjens:
I can agree with your reasoning in most of the cases. For the others, see my comments there.

I agree not all IDE recommendations may be valid and I guess it is better for me to stick with the parts of the code that I am actually using/changing since I do not know enough about the project yet. I appreciate you reviewing them anyway since it gives me a better understanding of some conventions and choices within the project.

elki/src/main/java/de/lmu/ifi/dbs/elki/algorithm/clustering/kmeans/initialization/PredefinedInitialMeans.java
kno10: Yes, the TODO is "do we need to clone here, or is it always safe to reuse". I'm not sure if the initial-cluster-centers API has been well enough specified. I believe most algorithms assume they are allowed to modify the double[] arrays at will; and most initializations will create a fresh array, too. But there could be some unintended side effects.

PR #24, opened 2016-05-02
Added CBLOF algorithm

patrickkostjens:
This is my implementation of the CBLOF algorithm by He et al.
I used the LOF and KMeansOutlier algorithms as the main examples for the implementation.
Furthermore, I put 0.7.2 as the `@since` version in the comments, but I am unsure whether that version is correct.

### DIFF
--- a/elki/src/main/java/de/lmu/ifi/dbs/elki/algorithm/outlier/clustering/CBLOF.java
+++ b/elki/src/main/java/de/lmu/ifi/dbs/elki/algorithm/outlier/clustering/CBLOF.java
@@ -76,7 +76,6 @@ import java.util.List;
  * </p>
  *
  * @author Patrick Kostjens
- * @since 0.7.2
  * 
  * @param <O> the type of data objects handled by this algorithm
  */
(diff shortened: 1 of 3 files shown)


kno10:
I think this could be modified to use `PrototypeModel` instead of `MeanModel` (to allow PAM etc. instead of k-means; and arbitrary distance functions), or it should allow `NumberVectorDistanceFunction` only instead of arbitary distances.
`@since` should be added when preparing a new release, on everything new in that release. `@Alias` is for backwards compatibility with old class names, so this is also not needed on new functionality.
`clusterBoundary` should probably be a `int`.
But this looks pretty good, I will merge it when I have time (I'm busy with other projects right now, sorry)

patrickkostjens:
I incorporated all except your first comment since I am not entirely sure how to make the modifications for that one. I agree with you that it would be better to use `PrototypeModel` instead of `MeanModel`. However, the `PrototypeModel` requires a type paramater that is different from the type paramater of the database and the CBLOF algorithm as it currently is. To be able to use the distance function of the database, the type of the `PrototypeModel` has to match that of the `DistanceQuery`, but I do not think that is possible since that is fixed to be the same as the database and algorithm parameter.

I hope this makes sense, but what I think it comes down to is that I do not know a way to perform distance queries on an arbitrary output type returned as the prototype by `PrototypeModel`.

kno10:
Merged in 512af8454e031339eee04193c6ab0b4769203bb3

PR #27, opened 2016-10-14
added association rule mining to Elki

fredps:
Added the package associationrulemining with an Algorithm for mining association rule and various interestingness measures.

### DIFF
--- a/elki/src/main/java/de/lmu/ifi/dbs/elki/algorithm/associationrulemining/AssociationRule.java
+++ b/elki/src/main/java/de/lmu/ifi/dbs/elki/algorithm/associationrulemining/AssociationRule.java
@@ -1,0 +1,160 @@
+package de.lmu.ifi.dbs.elki.algorithm.associationrulemining;
+/*
+ This file is part of ELKI:
+ Environment for Developing KDD-Applications Supported by Index-Structures
+
+ Copyright (C) 2016
+ Ludwig-Maximilians-Universität München
+ Lehr- und Forschungseinheit für Datenbanksysteme
+ ELKI Development Team
+
+ This program is free software: you can redistribute it and/or modify
+ it under the terms of the GNU Affero General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU Affero General Public License for more details.
+
+ You should have received a copy of the GNU Affero General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import de.lmu.ifi.dbs.elki.algorithm.itemsetmining.Itemset;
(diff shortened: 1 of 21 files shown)


elki/src/main/java/de/lmu/ifi/dbs/elki/algorithm/associationrulemining/AssociationRule.java
```diff
@@ -0,0 +1,160 @@
...
+  
+  /**
+   * Append to a string buffer.
+   * 
+   * @param buf Buffer
+   * @param meta Relation metadata (for labels)
+   * @return String buffer for chaining.
+   */
+  public StringBuilder appendTo(StringBuilder buf, VectorFieldTypeInformation<BitVector> meta){
+    buf.append(this.getAntecedent().appendTo(new StringBuilder(), meta));
```
kno10: Why not simply `getAntecedent().appendTo(buf, meta);`? Why a new `StringBuilder`?

elki/src/main/java/de/lmu/ifi/dbs/elki/algorithm/associationrulemining/AssociationRuleGeneration.java
```diff
@@ -0,0 +1,226 @@
...
+   * 
+   * @author Frederic Sautter
+   *
+   * @param itemset Sparse representation of an itemset.
+   * @param subsets List to fill with subsets
+   * @param pruning Array for pruning
+   * @param map HashMap where the itemsets are stored
+   */
+  private static void calcSubsets(int[] itemset, List<Itemset> subsets, boolean[] pruning, HashMap<Itemset, Itemset> map) {
+    long creator = (long) Math.pow(2, itemset.length) - 2;
```
kno10: Use integer math (bit shifting = power of two) rather than floating point. Why `-2`?

elki/src/main/java/de/lmu/ifi/dbs/elki/algorithm/itemsetmining/DenseItemset.java
```diff
@@ -68,6 +74,10 @@ public boolean containedIn(BitVector bv) {
   public long[] getItems() {
     return items;
   }
+  
+  public int[] toSparseRep() {
+    return new int[0];
```
kno10: Rather throw an exception than returning an incorrect value!

elki/src/main/java/de/lmu/ifi/dbs/elki/algorithm/itemsetmining/Itemset.java
```diff
@@ -76,6 +97,13 @@ public int getSupport() {
    * @return Itemset contents.
    */
   abstract public long[] getItems();
+  
+  /**
+   * Get a SparseItemset like representation of the Itemset.
+   * 
+   * @return Itemset contents.
+   */
+  abstract public int[] toSparseRep();
```
kno10: Rather than always forcing everything into `int[]`, it would be better to work with the `Itemset` objects directly, similar to the `SparseNumberVector` API, which is very memory efficient.

kno10:
I have omitted the changes to the Itemset classes, and rewritten rule generation completely. The branch is now merged via cherry-picking.

PR #35, opened 2017-08-07
Eliminated duplicate code within IntegerMinMax class. Deprecated code was left intact.

commandini:


kno10:
A similar code also exists in DoubleMinMax. Fixed both in e2585b195dac8204a8bfdc4d00a41b755de85c80

PR #37, opened 2017-08-12
Simplified CASHInterval#compareTo() method. Changed code to favoring usage of 'instanceof'.

commandini:
instanceof operator already returns false if the object is null.

https://stackoverflow.com/a/2950415


kno10:
Two issues here:

1. You removed the `obj == this || <full equals check>` optimization in several places.

2. `obj instanceof HardcodedClassName` is *not* the same as `this.getClass() != obj.getClass()` because of subclassing. Thus, if someone subclasses one of these classes, the `equals()` methods with your modifications will produce wrong results. In your pull request, consider
```
LPNormDistanceFunction a = new LPNormDistanceFunction(3);
WeightedLPNormDistanceFunction b=new WeightedLPNormDistanceFunction(3,new double[]{2});
```
In your code `a.equals(b)` (but not `b.equals(a)`)?
Because `b instanceOf LPNormDistanceFunction`.
But `a.getClass() != b.getClass()`, and they are *not* equal.

There is a good reason why Eclipse generates `equals` methods using `getClass` - this can lead to some really subtle and hard to find bugs (but yes, there are also cases where the stricter same-class requirement can be undesired). In my opinion, `instanceOf` in `equals` is only safe in `final` classes.

commandini:
@kno10 Thanks for the explanation. :+1:

PR #63, opened 2019-06-28
Avoid NPE when calling nextRandom on an estimated distribution

jcaesar:


kno10:
`new Random()` hidden in a class is almost always a bad idea in the context of data analysis, as you may want to have full control over the random numbers. And since the setter introduced only checks for null if `-ea` assertions are used, it does not achieve the desired goal of preventing this kind of NPE.

**The NPE is *fine*, in my opinion.**

It arises from programming errors, namely uninitialized use of random generation. Since the rng is `null`, the NPE is the appropriate kind of exception to throw. These errors *should* cause an exception, and should be fixed where they arise, not worked around by falling back to creating a new random generator hidden in an abstract parent class. This is more likely to cause unexpected results when people try to seed random generators and miss this hidden one.

It would be okay to add appropriate `assert`s with explanatory messages that help fixing this problem when it arises. So `nextRandom()` should get an `assert random != null : "If you want to use nextRandom(), pass a random generator to the constructor."` or similar.

jcaesar:
> `pass a random generator to the constructor.`

That isn't quite possible if the `Distribution` is constructed by an estimator, so I'd prefer to keep the setter. Otherwise, I'll modify that PR accordingly.

kno10:
Oh, I see. I'm usually not a big fan of setters but would prefer variables to be `final`, but here it may be okay.

kno10:
I am wondering whether it wouldn't be better to modify the API to not have a `Random` inside the distribution at all, but rather pass it as parameter via `nextRandom(Random rng)`. This would allow dropping the abstract class completely, and only have the interface, hence simplify the type hierarchy.

jcfj:
This does solve my problem, thank you. (And it's not exactly an uncommon pattern, c.f. e.g. the [C++ STL](http://www.cplusplus.com/reference/random/geometric_distribution/operator()/).)

PR #65, opened 2019-09-09
Implemented Bivariate Monte Carlo Density Estimation using Mann-Whitney-P Test (MCDE MWP)

alanmazankiewicz:
Algorithm is based on this paper: https://dbis.ipd.kit.edu/download/MCDE19_FOUCHE_SSDBM19.pdf

A abstract class `MCDEDependenceMeasure` has been created as subclass of `AbstractDependenceMeasure`. `MCDEDependenceMeasure` implements the generic Monte Carlo runs in the `dependence()` method returning the overall result of the algorithm. Further, it implements the generic slicing process and declares some methods (computation of statistical test and index structure) to be implemented by subclasses. Generally the slicing and the final MC runs are independent of the statistical test to be used. Each subclass should therefore only implement the statistical test (returning the p-value not the test statistic) and an appropriate index structure to efficiently compute the given statistical test. As each statistical test may need specific addition values beyond an index value and an adjustment for ties using `ranks()` from `AbstractDependenceMeasure` may not be enough. MWP for instance needs an additional correction value. In order to hold the data of the index structure tuple-like `RankStruct` and `MwpIndex` have been created in `utils/containers`. 

`McdeMwpDependenceMeasure` is a subclass of `MCDEDependenceMeasure` implementing MCDE using MWP as the test statistic. 

We decided on this structure instead of passing external statistical test classes (`GoodnessOfFitTest`) like in `HiCSDependenceMeasure` because in this way it is possible to use tailored index structures to more efficiently compute the resulting statistical tests.

Appropriate Test Classes have been created.


elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/McdeMwpDependenceMeasure.java
```diff
@@ -0,0 +1,227 @@
...
+            super.makeOptions(config);
+
+            final IntParameter mP = new IntParameter(M_ID, 50) //
+                    .addConstraint(CommonConstraints.GREATER_THAN_ONE_INT);
+            if(config.grab(mP)) {
+                m = mP.intValue();
+            }
+
+            final DoubleParameter alphaP = new DoubleParameter(ALPHA_ID, 0.5) //
+                    .addConstraint(CommonConstraints.GREATER_THAN_ZERO_DOUBLE); // actually > 0 and < 1 but no such constrain available
```
kno10: You can call `.addConstraint` multiple times to also add an upper bound.

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,173 @@
+package de.lmu.ifi.dbs.elki.math.statistics.dependence;
```
kno10: Can you please add the usual ELKI license (AGPL-3), so this code can be merged?

kno10: In the current master branch, we have simplified the package hierarchy from `de.lmu.ifi.dbs.elki` to just `elki`. It would be best if you can rebase your pull request on the current master branch.

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/McdeMwpDependenceMeasure.java
```diff
@@ -0,0 +1,227 @@
...
+        /**
+         * Random generator.
+         */
+        protected RandomFactory rnd;
+
+        @Override
+        protected void makeOptions(Parameterization config) {
+            super.makeOptions(config);
+
+            final IntParameter mP = new IntParameter(M_ID, 50) //
```
kno10: Parameters such as M_ID should usually go to the class where they are used, which here is the abstract parent class; and then should be inherited via the `super.makeOptions();` call. This way they won't be duplicated in multiple subclasses.

elki/src/main/java/de/lmu/ifi/dbs/elki/utils/containers/MwpIndex.java
```diff
@@ -0,0 +1,18 @@
+package de.lmu.ifi.dbs.elki.utils.containers;
+
+/**
+ * Structure to hold return values in index creation for McdeMwpDependenceMeausre
+ */
+
+public class MwpIndex extends RankStruct{
+    public double adjusted;
```
kno10: public fields (or ideally, any field) should come with a JavaDoc documentation. It may be appropriate to make this a static inner class of McdeMwpDependenceMeasure, if this is not useful independently of it.

elki/src/main/java/de/lmu/ifi/dbs/elki/utils/containers/RankStruct.java
```diff
@@ -0,0 +1,15 @@
+package de.lmu.ifi.dbs.elki.utils.containers;
+
+/**
+ * Structure to hold return values in index creation for MCDEDependenceEstimate
+ */
+
+public class RankStruct {
+    public int index;
```
kno10: This is essentially a boxed integer - a fairly costly object due to Java object memory management. We usually avoid small objects that occur in large quantities because of these overheads.

elki/src/test/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/McdeMwpDependenceMeasureTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Test
+    public void runTest() {
+
+        // independent
+        double[] indi1 = new double[1000];
+        double[] indi2 = new double[1000];
+
+        for (int i = 0; i < 1000; i++) {
+            indi1[i] = Math.random();
```
kno10: Please always use fixed random seeds, to avoid random test failures.

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/McdeMwpDependenceMeasure.java
```diff
@@ -0,0 +1,227 @@
...
+                n1++;
+            }
+        }
+
+        // This is to cancel the offset in case the marginal restriction does not start from 0
+        // see "acc - (cutStart * count)" is reference implementation of MWP
+        R -= start * n1;
+
+        final int cutLength = end - start;
+        if((n1 == 0) || (n1 == cutLength)) return 1;
```
kno10: By the ELKI code style (please also use the provided Eclipse code formatter in the `/conventions` folder to have our usual indentation of 2 spaces), please do *not* omit the `{}` parentheses. It is a best practise to always use parentheses even with single-line statements to avoid programming errors, such as the infamous Apple "goto fail" bug, which would have been prevented by not omitting the braces (and, of course, reachability analysis).

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/McdeMwpDependenceMeasure.java
```diff
@@ -0,0 +1,227 @@
...
+
+        final double U = R - ((double)(n1 * (n1 - 1))) / 2;
+        final long n2 = cutLength - n1;
+
+        final long two_times_sqrt_max_long = 6074000999L;
```
kno10: Magic numbers should be defined as constants, and it should be easy to see how they were obtained - it may be well acceptable to compute the constant on class load as 2*(long)Math.sqrt(Long.MAX_VALUE).

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/McdeMwpDependenceMeasure.java
```diff
@@ -0,0 +1,227 @@
...
+ * @author Edouard Fouché
+ */
+
+@Reference(authors = "Edouard Fouché, Klemens Böhm", //
+        title = "Monte Carlo Density Estimation", //
+        booktitle = "Proc. 2019 ACM Int. Conf. on Scientific and Statistical Database Management (SSDBM 2019)",
+        url = "https://doi.org/10.1145/3335783.3335795", //
+        bibkey = "DBLP:conf/ssdbm/FoucheB19")
+
+public class McdeMwpDependenceMeasure extends MCDEDependenceMeasure<MwpIndex> {
```
kno10: Because MCDE is an abbreviation, wouldn't the class name `MCDEMWPDependenceMeasure` be more consistent?

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,173 @@
...
+     * Data Slicing
+     *
+     * @param len No of data instances
+     * @param nonRefIndex Index (see correctedRank()) computed for the dimension that is not the reference dimension
+     * @return Array of booleans that states which instances are part of the slice
+     */
+    protected boolean[] randomSlice(int len, R[] nonRefIndex){
+        final Random random = rnd.getSingleThreadedRandom();
+        boolean slice[] = new boolean[len];
+        Arrays.fill(slice, Boolean.TRUE);
```
kno10: For filling a primitive array just use `true`.

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,173 @@
...
+ */
+
+@Reference(authors = "Edouard Fouché, Klemens Böhm", //
+        title = "Monte Carlo Density Estimation", //
+        booktitle = "Proc. 2019 ACM Int. Conf. on Scientific and Statistical Database Management (SSDBM 2019)",
+        url = "https://doi.org/10.1145/3335783.3335795", //
+        bibkey = "DBLP:conf/ssdbm/FoucheB19")
+
+
+public abstract class MCDEDependenceMeasure<R extends RankStruct> extends AbstractDependenceMeasure {
```
kno10: I'm wondering if the abstract class is useful as is; because there is only one subclass, right?
This would allow eliminating the entire `RankStruct` construct? For long-term maintainability, I have been repeatedly trying to simplify class hierarchies, removing generics, etc.

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * Constructor.
+     *
+     * @param m Monte-Carlo iterations
+     * @param alpha Expected share of instances in slice (under independence)
+     * @param beta Share of instances in marginal restriction (reference dimension)
+     * @param rnd Random source
+     */
+    public MCDEDependenceMeasure(int m, double alpha, double beta, RandomFactory rnd){
+        if((beta > 1.0) || beta <= 0.0) throw new AbortException("beta must be in ]0;1]");
```
kno10: In ELKI we currently usually don't test these in the constructor, but only in the parameterizer with the constraints. It would likely be better to systematically add such parameter checks throughout ELKI (obviously, rnd should then also be tested to be not null though); but these shouldn't use the rather nondescript `AbortException` but rather `IllegalArgumentException` or similar.

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,173 @@
...
+        if((alpha >= 1.0) || alpha <= 0.0) throw new AbortException("alpha must be in ]0;1[ ");
```
kno10: The if condition is better phrased as `!(alpha > 0. && alpha < 1.0)` because of NaN values.

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,173 @@
...
+     * @param len No of data instances
+     * @param nonRefIndex Index (see correctedRank()) computed for the dimension that is not the reference dimension
+     * @return Array of booleans that states which instances are part of the slice
+     */
+    protected boolean[] randomSlice(int len, R[] nonRefIndex){
+        final Random random = rnd.getSingleThreadedRandom();
+        boolean slice[] = new boolean[len];
+        Arrays.fill(slice, Boolean.TRUE);
+
+        final int slizeSize = (int) Math.ceil(Math.pow(this.alpha, 1.0) * len);
```
kno10: `pow(x,1)` clearly is an error, maybe a leftover from an earlier version that used a different exponent?

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/McdeMwpDependenceMeasure.java
```diff
@@ -0,0 +1,227 @@
...
+            int k = j; int t = 1; double adjust = 0.0;
+
+            while((k < len - 1) && (adapter.getDouble(data, idx[k]) == adapter.getDouble(data, idx[k+1]))){
+                adjust += k;
+                k++; t++;
+            }
+
+            if(k > j){
+                double adjusted = (adjust + k) / t;
+                correction += (t*t*t) - t;
```
kno10: As far as I can tell, `correction` is always integer. These values here likely can get quite large - wouldn't a `long` be more appropriate? Also, the data is later stored as a `double`?

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/McdeMwpDependenceMeasure.java
alanmazankiewicz: Not relevant anymore

elki/src/main/java/de/lmu/ifi/dbs/elki/utils/containers/MwpIndex.java
alanmazankiewicz: RankStruct and MwpIndex are now inner classes of the new MCDETest and MWPTest accordingly

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/McdeMwpDependenceMeasure.java
alanmazankiewicz: Not relevant anymore

elki/src/main/java/de/lmu/ifi/dbs/elki/math/statistics/dependence/MCDEDependenceMeasure.java
alanmazankiewicz: Yes I adjusted it. We now have the abstract `MCDETest` and its subclass `MWPTest` that creates the index and computes the statistical test. They get passed to `MCDEDependenceMeasure` They also contain `RankStruct` and `MwpIndex` as inner classes. The reason I used inheritance is that the way we compute MWP is very tailored to the whole MCDE framework to be efficient therefore the tests are not really usefull outside of it. `McdeMwpDependenceMeasure` has been deleted.

elki/src/test/java/elki/math/statistics/dependence/McdeMwpDependenceMeasureTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    double[] more_noise_lin1 = new double[1000];
+    double[] more_noise_lin2 = new double[1000];
+
+    for(int i = 0; i < 1000; i++) {
+      more_noise_lin1[i] = i;
+      more_noise_lin2[i] = (i * 2) + generator.nextGaussian() * 500;
+    }
+
+    NumberArrayAdapter adapter = new ExtendedArray<Double>();
```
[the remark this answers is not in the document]
alanmazankiewicz: yes true

alanmazankiewicz:
Friendly reminder :)

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MCDETest.java
```diff
@@ -0,0 +1,93 @@
...
+ * See MWPTest for an example. Implementation should contain an appropriate index
+ * structure to efficiently compute statistical test and the test itself.
+ *
+ * @param <R> RankStruct or Subclass of RankStruct
+ *
+ * @author Alan Mazankiewicz
+ * @author Edouard Fouché
+ */
+
+public abstract class MCDETest<R extends MCDETest.RankStruct> {
```
kno10: We usually have no whitespace lines around the class declaration, but I can adjust these when merging.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MCDETest.java
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Structure to hold return values in index creation for MCDEDependenceEstimate
+   */
+  public class RankStruct {
```
kno10: should be a `static` inner class (saves some memory), and should have JavaDoc on attributes and methods.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MCDETest.java
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Build a sorted index of objects.
+   *
+   * @param adapter Data adapter
+   * @param data    Data array
+   * @param len     Length of data
+   * @return Sorted index
+   */
+  protected <A> int[] sortedIndex(final NumberArrayAdapter<?, A> adapter, final A data, int len) {
```
kno10: As this is only used in the 1-liner `corrected_ranks` method below, it could probably be inlined there?

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MCDETest.java
```diff
@@ -0,0 +1,93 @@
...
+    int[] s1 = MathUtil.sequence(0, len);
+    IntegerArrayQuickSort.sort(s1, (x, y) -> Double.compare(adapter.getDouble(data, x), adapter.getDouble(data, y)));
+    return s1;
+  }
+
+  /**
+   * Overloaded wrapper for corrected_ranks()
```
kno10: This JavaDoc is not very informative, and does not document parameters or return values (likely mostly a copy and paste from the other corrected_ranks below).

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MCDETest.java
```diff
@@ -0,0 +1,93 @@
...
+   */
+  public <A> R[] corrected_ranks(final NumberArrayAdapter<?, A> adapter, final A data, int len) {
```
kno10: Java conventions are to use CamelCase, i.e. `correctedRanks`. Underscores are python conventions.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Implementation of Mann-Whitney U test returning the p-value (not the test
+ * statistic, thus MWP) for MCDEDependenceMeasure. Implements algorithm 1 and 3
+ * of reference paper.
+ *
+ * @author Alan Mazankiewicz
+ * @author Edouard Fouché
+ */
+
+@Reference(authors = "Edouard Fouché, Klemens Böhm", //
```
kno10: Because `@Reference` annotates the `class`, there shouldn't be an empty line inbetween.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+    title = "Monte Carlo Density Estimation", //
+    booktitle = "Proc. 2019 ACM Int. Conf. on Scientific and Statistical Database Management (SSDBM 2019)", url = "https://doi.org/10.1145/3335783.3335795", //
+    bibkey = "DBLP:conf/ssdbm/FoucheB19")
+
+public class MWPTest extends MCDETest<MWPTest.MwpIndex> {
+
+  private static final long two_times_sqrt_max_long = 2 * (long) Math.sqrt(Long.MAX_VALUE);
```
kno10: Constants should be uppercase. This constant could be named by its meaning for the algorithm, not just the mathematical definition. For example `MAX_SAMPLES` or so. But I do not think this is actually needed, as explained below.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+   */
+  protected RandomFactory rnd;
+
+  /**
+   * Structure to hold values needed for computing MWP in MCDEDependeneMeasure.
+   * Values get computed during index creation.
+   */
+
+  public class MwpIndex extends MCDETest.RankStruct {
+    long correction;
```
kno10: Why is `correction` not public, but `adjusted` is?
The fields and methods should have JavaDoc documentation.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+   */
+
+  protected MWPTest(){
+  }
+
+  /**
+   * Computes Corrected Rank Index as described in Algorithm 1 of reference paper.
+   * The notation follows ELKI convention if applicable, else we use the notation from the reference paper.
+   *
+   * @param adapter ELKI NumberArrayAdapter Subclass
```
kno10: actually its an instance, not a subclass.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+      this.correction = correction;
+      this.adjusted = adjusted;
+    }
+  }
+
+  /**
+   * Static Constructor - use this one
+   */
+
+  public static final MWPTest STATIC = new MWPTest();
```
kno10: This should go to the very top of the class (constant).

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+   * Computes Corrected Rank Index as described in Algorithm 1 of reference paper.
+   * The notation follows ELKI convention if applicable, else we use the notation from the reference paper.
+   *
+   * @param adapter ELKI NumberArrayAdapter Subclass
+   * @param data    One dimensional array containing one dimension of the data
+   * @param idx     Return value of sortedIndex()
+   * @return Array of doubles, 3 subsequent values being assigned to one data instance.
+   * Containing sorted (ascending) row numbers, adjusted ranks and tying value corrections
+   * as required by MWP test. Example:
+   * double[] corrected_ranks = corrected_ranks(...);
```
kno10: Code examples should be wrapped in `{@code }` so they do not break when auto-formatted. The `@return` tag should have a short explanation, the usage example should go to the main JavaDoc of the method instead.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+      if(k > j) {
+        double adjusted = (adjust + k) / t;
+        correction += (t * t * t) - t;
+
+        for(int m = j; m <= k; m++) {
+          I[m] = new MwpIndex(idx[m], adjusted, correction);
+        }
+      }
+      else
```
kno10: Missing `{}`

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+    final int cutLength = end - start;
+    if((n1 == 0) || (n1 == cutLength)) {
+      return 1;
+    }
+
+    final double U = R - ((double) (n1 * (n1 - 1))) / 2;
+    final long n2 = cutLength - n1;
+
+    if(n1 + n2 > two_times_sqrt_max_long) {
```
kno10: By the definition of `n2`, `n1 + n2` always is `cutLength`. Because `cutLength` is integer, it will always be less than this constant?

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+    final long b_end = corrected_ranks[(end - 1)].correction;
+    final long b_start = start == 0 ? 0 : corrected_ranks[(start - 1)].correction;
+    final double correction = (double) (b_end - b_start) / (cutLength * (cutLength - 1));
+    final double std = Math.sqrt((((double) (n1 * n2)) / 12) * (cutLength + 1 - correction));
+
+    if(std == 0) {
+      return 0;
+    }
+    else {
```
kno10: The `else {` + `}` is not necessary, because of the `return` above for this special case.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+    if(n1 + n2 > two_times_sqrt_max_long) {
+      throw new AbortException("Long type overflowed. Too many objects: Please subsample and try again with smaller data set.");
+    }
+
+    final long b_end = corrected_ranks[(end - 1)].correction;
+    final long b_start = start == 0 ? 0 : corrected_ranks[(start - 1)].correction;
+    final double correction = (double) (b_end - b_start) / (cutLength * (cutLength - 1));
+    final double std = Math.sqrt((((double) (n1 * n2)) / 12) * (cutLength + 1 - correction));
+
+    if(std == 0) {
```
kno10: Can this actually happen? Maybe this case could be caught earlier (`n1==0 || n2 == 0`)?

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
+/*
```
kno10: This class should be moved to the `elki-core-math` module. The HICSDependenceMeasure is a special case, because it imports from the `HiCS` class, which is in the `elki-outlier` module, and we don't want a dependency from `elki-core-math` onto `elki-outlier` (but I'll try to resolve this by refactoring - most things in the `elki` main module should eventually be moved into a more appropriate place).

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+import elki.utilities.optionhandling.parameters.RandomParameter;
+import elki.utilities.random.RandomFactory;
+import elki.math.statistics.tests.mcde.*;
+
+import java.util.Arrays;
+import java.util.Random;
+
+/**
+ * Implementation of bivariate Monte Carlo Density Estimation as described in
+ * Edouard Fouché & Klemens Böhm<br>
```
kno10: The first sentence of a JavaDoc should be a short regular sentence (because it will be used as a short description of the class). We usually have the references at the end of the JavaDoc because of this.

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+
+  /**
+   * Random generator.
+   */
+  protected RandomFactory rnd;
+
+  /**
+   * Statistical Test returning p-value tailored to MCDE Framework.
+   */
+  protected MCDETest mcdeTest;
```
kno10: This line (and many more below) has a warning, because MCDETest has a generic.

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+
+  /**
+   * Constructor.
+   */
+  public MCDEDependenceMeasure(int m, double alpha, double beta, RandomFactory rnd, MCDETest mcdeTest) {
```
kno10: Parameters should be documented in the JavaDoc.

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+      throw new AbortException("Size of both arrays must match!");
+    }
+
+    final MCDETest.RankStruct[] index_0 = mcdeTest.corrected_ranks(adapter1, data1, len);
+    final MCDETest.RankStruct[] index_1 = mcdeTest.corrected_ranks(adapter2, data2, len);
+
+    double mwp = 0;
+
+    for(int i = 0; i < this.m; i++) {
+      int r = random.nextInt(2);
```
kno10: `boolean flip = random.nextInt(2) == 1;`

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+      MCDETest.RankStruct[] ref_index;
```
kno10: `ref_index = flip ? index_1 : index_0;`
`other_index = flip ? index_0 : index_1;`
Some people have issues with ternary statements, but they often are much more concise and usually produce shorter bytecode and even faster code in the end. Alternatively, the case distinction could be moved below `start` and `end` so these two variables are not used at all.

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+      }
+
+      final int start = random.nextInt((int) (len * (1 - this.beta)));
+      final int end = start + (int) Math.ceil(len * this.beta);
+      mwp += mcdeTest.statistical_test(start,end, randomSlice(len, other_index), ref_index);
+    }
+    return mwp / m;
+  }
+
+  public static class Par implements Parameterizer {
```
kno10: While the JavaDoc of `Par` classes is totally redundant, we try to have JavaDoc everywhere.

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+     * Random generator.
+     */
+    protected RandomFactory rnd;
+
+    @Override public void configure(Parameterization config) {
+      new IntParameter(M_ID, 50) //
+          .addConstraint(CommonConstraints.GREATER_THAN_ONE_INT).grab(config, x -> m = x);
+
+      new DoubleParameter(ALPHA_ID, 0.5) //
+          .addConstraint(CommonConstraints.GREATER_THAN_ZERO_DOUBLE).addConstraint(CommonConstraints.LESS_THAN_ONE_DOUBLE).grab(config, x -> alpha = x);
```
kno10: these should be multiple lines (use ` //` to prevent the code formatter to join the lines).

elki/src/test/java/elki/math/statistics/dependence/MCDEDependenceMeasureTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+public class MCDEDependenceMeasureTest {
+
+  @Test public void runTest() {
+
+    // independent
+    double[] indi1 = new double[1000];
+    double[] indi2 = new double[1000];
+
+    RandomFactory rnd = new RandomFactory(0);
```
kno10: In unit tests, because these are not parameterizable, it is fine to use `new Random(0L)` to fix the random seed.
`RandomFactory` is a construct aimed at making it easier to globally fix the random seeds for benchmarking.

elki/src/test/java/elki/math/statistics/dependence/MCDEDependenceMeasureTest.java
```diff
@@ -0,0 +1,138 @@
...
+    for(int i = 0; i < 1000; i++) {
+      more_noise_lin1[i] = i;
+      more_noise_lin2[i] = (i * 2) + generator.nextGaussian() * 500;
+    }
+
+    DoubleArrayAdapter adapter = DoubleArrayAdapter.STATIC;
+    MWPTest mwpTest = MWPTest.STATIC;
+
+    MCDEDependenceMeasure mwp = new ELKIBuilder<>(MCDEDependenceMeasure.class) //
+        .with(M_ID, 1000)
```
kno10: For consistency, please use `MCDEDependenceMeasure.Par.M_ID` instead of the static import. Also add ` //` to the end of the lines to prevent automatic code formatting to join them.

elki/src/test/java/elki/math/statistics/dependence/MCDEDependenceMeasureTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    test_result(indi1, indi2, 0.65, 0.35, 0.67, 0.33, adapter, mwp);
+    test_result(lin1, lin2, 1.0, 0.99, 1.0, 0.97, adapter, mwp);
+    test_result(small_noise_lin1, small_noise_lin2, 1.0, 0.96, 1.0, 0.85, adapter, mwp);
+    test_result(more_noise_lin1, more_noise_lin2, 1.0, 0.85, 1.0, 0.83, adapter, mwp);
+
+    test_rankIndex(adapter, mwpTest);
+  }
+
+  public static void test_result(double[] data1, double[] data2, double strict_upper, double strict_lower, double lax_upper, double lax_lower, NumberArrayAdapter adapter, MCDEDependenceMeasure mwp) {
```
kno10: Java conventions are to use CamelCase for method names. `testResult` is a bit nondescript, maybe `testTolerance`?

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
kno10: The easiest solution that I could find is to type this as `MCDETest<MCDETest.RankStruct>`, and perform a single cast in `MCDEDependenceMeasure`. However, this may - even as it is now - cause some subtle class cast exception when someone actually extends this later.

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+   */
+  @Override public <A, B> double dependence(final NumberArrayAdapter<?, A> adapter1, final A data1, final NumberArrayAdapter<?, B> adapter2, final B data2) {
+    final Random random = rnd.getSingleThreadedRandom();
+    final int len = adapter1.size(data1);
+
+    if(len != adapter2.size(data2)) {
+      throw new AbortException("Size of both arrays must match!");
+    }
+
+    final MCDETest.RankStruct[] index_0 = mcdeTest.corrected_ranks(adapter1, data1, len);
```
kno10: This array type is likely to break in the future. An `ArrayList<MCDETest.RankStruct>` is more tolerant, because then the actual type is erased in the compiled classes - its then just an `ArrayList` eventually.
Now if `corrected_ranks` would return a `MWPTest.MwpIndex[]`, this could (should?) cause an error, because of type covariance. The warning caused by the missing generics is actually hinting at such a problem, and I am actually surprised that it works.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+   * Containing sorted (ascending) row numbers, adjusted ranks and tying value corrections
+   * as required by MWP test. Example:
+   * double[] corrected_ranks = corrected_ranks(...);
+   * double l = corrected_rank[0]; double adjusted_rank = corrected_rank[1]; double correction = corrected_rank[2];
+   * <p>
+   * Correspond to one instance of the original data
+   */
+  public <A> MwpIndex[] corrected_ranks(final NumberArrayAdapter<?, A> adapter, final A data, int[] idx) {
+    final int len = adapter.size(data);
+    MwpIndex[] I = (MwpIndex[]) Array.newInstance(MwpIndex.class, len);
```
kno10: `new MwpIndex[len];` no generics involved.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+    final double correction = (double) (b_end - b_start) / (cutLength * (cutLength - 1));
+    final double std = Math.sqrt((((double) (n1 * n2)) / 12) * (cutLength + 1 - correction));
+
+    if(std == 0) {
+      return 0;
+    }
+    else {
+      final double mean = ((double) (n1 * n2)) / 2;
+      final double Z = Math.abs((U - mean) / std);
+      return erf(Z / Math.sqrt(2)); // Note that this is equivalent to do 1-2*(1-cdf(Z,0,1));
```
kno10: Here I would prefer `NormalDistribution.erf`.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+   * Random generator.
+   */
+  protected RandomFactory rnd;
+
+  /**
+   * Structure to hold values needed for computing MWP in MCDEDependeneMeasure.
+   * Values get computed during index creation.
+   */
+
+  public class MwpIndex extends MCDETest.RankStruct {
```
kno10: `MWPIndex` uppercase, for consistency with the wrapping class

kno10:
I will likely have to merge this as a single commit, because rebasing causes some conflict in almost every merge (likely because of the merges that were performed before). They all seem trivial, but will require some work. But I guess that is okay.

I have given you another round of detailed comments. Some of them I could trivially have changed when merging; like code style. The biggest thing to resolve is that generics warnings.

Would it make sense to change `MCDETest.RankStruct[]` into a `MCDETest.RankStruct` that stores an `int[]` instead (this would save a lot of memory, and be faster)?
Similarly, the subtype `MWPTest.MwpIndex` could store a `long[] correction` and `double[] adjusted`. So instead of N objects (at 64+ bytes each), this would be one such object for each attribute, I guess. This should use some 16-24 x N bytes less memory, and will likely be faster to access, too.

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+    return mwp / m;
+  }
+
+  public static class Par implements Parameterizer {
+    /**
+     * Parameter that specifies the number of iterations in the Monte-Carlo
+     * process of identifying high contrast subspaces.
+     */
+    public static final OptionID M_ID = new OptionID("MCDE.test", "No. of Monte-Carlo iterations.");
```
kno10: The option ID should be "mcde.m".

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * Parameter that specifies the size of the slice
+     */
+    public static final OptionID ALPHA_ID = new OptionID("MCDE.test", "Expected share of instances in slice (independent dimensions).");
```
kno10: The option ID should be "mcde.alpha".

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * Parameter that specifies the size of the marginal restriction. Note that in the original paper
+     * alpha = beta and as such there is no explicit distinction between the parameters.
+     */
+    public static final OptionID BETA_ID = new OptionID("MCDE.test", "Expected share of instances in marginal restriction (dependent dimensions).");
```
kno10: The option ID should be "mcde.beta".

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * Parameter that specifies the random seed.
+     */
+    public static final OptionID SEED_ID = new OptionID("MCDE.test", "The random seed.");
```
kno10: The option ID should be "mcde.seed".

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * Computes Corrected Rank Index as described in Algorithm 1 of reference paper.
+   * The notation follows ELKI convention if applicable, else we use the notation from the reference paper.
+   *
+   * @param adapter ELKI NumberArrayAdapter Subclass
+   * @param data    One dimensional array containing one dimension of the data
+   * @param idx     Return value of sortedIndex()
+   * @return Array of doubles, 3 subsequent values being assigned to one data instance.
+   * Containing sorted (ascending) row numbers, adjusted ranks and tying value corrections
```
kno10: "tie", "tieing"

elki/src/test/java/elki/math/statistics/dependence/MCDEDependenceMeasureTest.java
kno10: Also, since data1 and data2 are double[], there is no benefit in making the adapter a parameter. Its always the standard double[] adapter.

elki/src/test/java/elki/math/statistics/dependence/MCDEDependenceMeasureTest.java
```diff
@@ -0,0 +1,138 @@
...
+      total_res += res;
+      assertTrue("MWP result out of acceptable range. Result: " + res + ". Lax upper bound: " + lax_upper, (res <= lax_upper));
+      assertTrue("MWP result out of acceptable range. Result: " + res + ". Lax lower bound: " + lax_lower, (res >= lax_lower));
+    }
+    total_res /= 100;
+    assertTrue("MWP result out of acceptable range. Result: " + total_res + ". Strict upper bound: " + strict_upper, (total_res <= strict_upper));
+    assertTrue("MWP result out of acceptable range. Result: " + total_res + ". Strict lower bound: " + strict_lower, (total_res >= strict_lower));
+  }
+
+  public static void test_rankIndex(NumberArrayAdapter adapter, MWPTest mwpTest) {
```
kno10: Since this tests MWPTest, it should go into a separate class MWPTestTest

elki/src/test/java/elki/math/statistics/dependence/MCDEDependenceMeasureTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    DoubleArrayAdapter adapter = DoubleArrayAdapter.STATIC;
+    MWPTest mwpTest = MWPTest.STATIC;
+
+    MCDEDependenceMeasure mwp = new ELKIBuilder<>(MCDEDependenceMeasure.class) //
+        .with(M_ID, 1000)
+        .with(TEST_ID, mwpTest)
+        .build();
+
+    test_result(indi1, indi2, 0.65, 0.35, 0.67, 0.33, adapter, mwp);
```
kno10: These are four tests wrapped in one. This is better broken into four separate tests (so the log will show which one failed).

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+
+    // According to the actual formula it should be Math.ceil(Math.pow(this.alpha, 1.0) * len).
+    // The exponent in the multivariate case should be the no. of dimensions - 1 which in the
+    // bivariate case is always simply 1.
+    final int slizeSize = (int) Math.ceil(this.alpha * len);
+    final int start = random.nextInt(len - slizeSize);
+    final int end = start + slizeSize;
+
+    for(int j = 0; j < start; j++) {
+      slice[nonRefIndex[j].index] = false;
```
kno10: Wouldn't it make sense to iterate start..end, and only set these to true, rather than first filling everything with true, then setting those to false?

elki/src/main/java/elki/math/statistics/dependence/MCDEDependenceMeasure.java
```diff
@@ -0,0 +1,249 @@
...
+   *
+   * @param adapter1 First data adapter
+   * @param data1    First data set
+   * @param adapter2 Second data adapter
+   * @param data2    Second data set
+   * @param <A>      Numeric data type, such as double
+   * @param <B>      Numeric data type, such as double
+   * @return MCDE result
+   */
+  @Override public <A, B> double dependence(final NumberArrayAdapter<?, A> adapter1, final A data1, final NumberArrayAdapter<?, B> adapter2, final B data2) {
```
kno10: Note that the interface `DependenceMeasure` also includes a pair-wise API. It would be more efficient to build the indexes only once each.

kno10:
I have pushed an edited version to branch https://github.com/elki-project/elki/tree/feature/mcde
This is not completely done yet, but maybe you can have a look at the edits I made.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
alanmazankiewicz: This mechanism was supposed to secure the user against using too many instances resulting in `n1*n2` being too large to fit into a `long` and thus resulting in a negative number which in turn would result in returning `NaN` after the MCDE runs. The user then does not understand why `NaN` is returned because he assumes that everything he does is correct. I have taken over this from out Scala implementation where it was tested and worked well. However from looking at the code I see your point as `cutLength` as well as `start` and `end` are all ints. I will look at it more closely soon and get back to you about it if necessary.

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
alanmazankiewicz: Obviously it is important to not divide by 0 in line 178. Neither n1 nor n2 should get 0 so the only way this can get 0 is if correction == cutLength + 1. Today I could not really create a situation in which that happens and the only scenario I could intuitively imagine is if there is no variation in the data which can be ruled out as I tested it and it lead not to std being 0. I will talk with Eduard about it as he came up with this way of calculating the standard deviation and if necessary come back to you about it.

alanmazankiewicz:
Thank you so much for taking care of those changes. I looked at the code, tested it and merged it into my master as it seems to work. I just had to take care of a comment at one point. Let me know if you do any further changes. Please note the my two comments above.

kno10:
I was wondering if you want to also add the "higher order" version, maybe as `public <A> double higherOrderDependence(NumberArrayAdapter<?, A> adapter, List<? extends A> data)` for completeness. There wouldn't be alternatives for this use case right now (and its not in the DependencyMeasure interface), but someone searching for MCDE code and looking at ELKI would still find it.
Thanks for the fix to the JavaDoc, I will add this.
After committing this to master, I will delete this feature/mcde branch again. Afterwards I will also rename all the `*DependencyMeasure` classes to just `Dependency`, and in some cases I may also omit the `Dependency` part; just to make class names shorter everywhere. The `Measure` is a bit redundant, and unless there is a name collision (HiCS for example), I will now prefer shorter names. So `DistanceCorrelationDependencyMeasure` will become just `DistanceCorrelation` (its widely known as dCor or dCor², but that does not make a good class name). Maybe I will even use DCor. Because in the end "distance" does suit this very well, it uses one-dimensional differences.

alanmazankiewicz:
What you mean by higher order version? As far as I can tell you have overridden a `dependence` method having the same signature `public <A> double dependence(NumberArrayAdapter<?, A> adapter, List<? extends A> data)`.

kno10:
The method I added produces *pairwise* bivariate results. I think your method is intended to make higher-order correlations of multiple variates, right (not sure how they are defined though; whether it is a 1 with d-1 dimensions correlation as in HiCS?) Essentially the case where you *do* need the pow(alpha or beta, dim-1)

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
alanmazankiewicz: I tested it and it is possible for `std` to be 0. What I did is I created a small number of random data (int) in a small range (0-10) (to enforce situations on which all data points in a slice are of the same value etc.) and let the parameter alpha and beta vary (so that the two slices also become uneven). As such I would suggest to add back 

```suggestion
    if(std == 0) {
    return 0; 
    }
```

The code of the test is: 
```suggestion
package elki.math.statistics.tests.mcde;

import elki.math.statistics.dependence.MCDEDependenceMeasure;
import elki.utilities.random.RandomFactory;

import java.util.Random;

public class Test {
  public static void main(String[] args){
    MWPTest mwp = MWPTest.STATIC;
    final int len = 10;
    final int value_bound = 10;

    double[] data_1 = new double[len];
    double[] data_2 = new double[len];
    RandomFactory rnd = new RandomFactory(0);
    final Random random = rnd.getSingleThreadedRandom();

    for(int j = 0; j < 10000; j++) {
      for(int i = 0; i < len; i++){
        data_1[i] = random.nextInt(value_bound);
        data_2[i] = random.nextInt(value_bound);
      }

      double beta = 0.1 + 0.9 * random.nextDouble();
      double alpha = 0.1 + 0.9 * random.nextDouble();

      MCDEDependenceMeasure mcde = new MCDEDependenceMeasure(1000, alpha, beta, rnd, mwp);
      mcde.dependence(data_1, data_2);
    }
  }
}
```

If you set an assertion or a conditional breakpoint in `MWPTest.statisticalTest()` you should be able to recreate it. Note that I have just created a `Test.java` file in the mcde [folder](https://github.com/elki-project/elki/tree/feature/mcde/elki-core-math/src/main/java/elki/math/statistics/tests/mcde) (where also `MCDETest` and `MWPTest` lie).

alanmazankiewicz:
https://github.com/alanmazankiewicz/elki/blob/c17ac459c81d9872892f3fb823350da91ce57e9e/elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java#L58-L63

Btw. I have just seen that those lines of code are not necessary anymore, it is a relict from an older version.

alanmazankiewicz:
> The method I added produces _pairwise_ bivariate results. I think your method is intended to make higher-order correlations of multiple variates, right (not sure how they are defined though; whether it is a 1 with d-1 dimensions correlation as in HiCS?) Essentially the case where you _do_ need the pow(alpha or beta, dim-1). Sorry, I have not yet read your paper, so I do not use the exact same terminology as you used.

Yes! I was gonna ask you if you could provide us with an abstract method for computing multivariate dependences so that we can implement a multivariate version of MCDE. I will implement it using this signature `public <A> double higherOrderDependence(NumberArrayAdapter<?, A> adapter, List<? extends A> data)` simply in `MCDEDependenceMeasure`. However, it would probably be best if we first get this pull request through (as you are going to do some further changes to the classnames etc.) and then I will  fetch elki again and take care of the new method?

elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java
alanmazankiewicz: I just learned that arrays in Java cannot be larger than `Integer.MAX_VALUE`. In that case I would suggest to simply convert  `n1` and `n2` into a `double`, then there should be no issue with `n1*n2` exceeding the max limit of the data type. Note that the double cast when computing `std` will not be required anymore.

alanmazankiewicz:
One more thing I just found. The correction value of the `MWPRanking` has to be `double[]`.
I checked it and in following lines the change has to be made: 

https://github.com/elki-project/elki/blob/b183fa6028e429ec0eac65fb5ef6fe2994e70bb6/elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java#L80

https://github.com/elki-project/elki/blob/b183fa6028e429ec0eac65fb5ef6fe2994e70bb6/elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java#L94

https://github.com/elki-project/elki/blob/b183fa6028e429ec0eac65fb5ef6fe2994e70bb6/elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java#L112

https://github.com/elki-project/elki/blob/b183fa6028e429ec0eac65fb5ef6fe2994e70bb6/elki-core-math/src/main/java/elki/math/statistics/tests/mcde/MWPTest.java#L173-L175
(in L175 the double cast is not required anymore)

kno10:
Why a double? The values accumulated in correction are all integers. Yes, if you have many of them they can overflow, but I'd rather catch this than use a double here. Because if this long overflows, the double will run out of precision much earlier, and accumulation will lose all small values and become unreliable, too.

alanmazankiewicz:
Yes you are right, correction can stay a long. When I was investigating if std == 0 is possible (see responds above btw.) I just noticed that our scala implementation uses double. Regarding the overflow I trust you to handle this appropriately however you see fit.

kno10:
Closing, as the rebased version has been merged.

PR #66, opened 2019-10-24
Created Multivariate MCDE

alanmazankiewicz:
- Implemented multivariate MCDE as `MCDEDependence.higherOrderDependence()` and additional `MCDEDependence.randomSlice()`. For the multivariate `randomSlice()` it is important to initialise the boolean array with `true` and then "false" out with the two loops due to looping over multiple dimensions (otherwise we get false results). 

- Implemented `MWPTest.safeCut()`. Prevents slicing in the middle of ties which may lead to errors / false results in some situations. For this `statisticalTest()` receives the `width` value instead of `end` and `RankStruct` has to hold a `len` value. 

- Deleted passing `len` to `randomSlice()`

- Added back` if(std == 0) return 0;` as this situation may arise in `statisticalTest()`

- Implemented additional test for `statisticalTest()`


elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MWPTest.java
```diff
@@ -100,14 +106,14 @@ private MWPTest() {
...
   public <A> MWPRanking correctedRanks(NumberArrayAdapter<?, A> adapter, A data, int len) {
-    int[] idx = Dependence.Util.sortedIndex(adapter, data, len);
     // Note: As described in Algorithm 1 of reference paper.
-    MWPRanking I = new MWPRanking(idx);
+    int[] idx = MathUtil.sequence(0, len);
+    IntegerArrayQuickSort.sort(idx, (x, y) -> Double.compare(adapter.getDouble(data, x), adapter.getDouble(data, y)));
+    MWPRanking I = new MWPRanking(idx, len);
     long correction = 0;
     for(int j = 0; j < len;) {
-      final double v = adapter.getDouble(data, idx[j]);
```
kno10: This construct was inserted for efficiency.

elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MWPTest.java
```diff
@@ -139,53 +142,81 @@ private MWPTest() {
...
     final double U = R - 0.5 * n1 * (n1 - 1);
     final int n2 = cutLength - n1;
-    final long b_end = corrected_ranks.correction[end - 1];
-    final long b_start = start == 0 ? 0 : corrected_ranks.correction[start - 1];
+    final long b_end = corrected_ranks.correction[safeEnd - 1];
+    final long b_start = safeStart == 0 ? 0 : corrected_ranks.correction[safeStart - 1];
     final double correction = (double) (b_end - b_start) / (cutLength * (cutLength - 1));
     final double std = FastMath.sqrt((((double) (n1 * n2)) / 12) * (cutLength + 1 - correction));
-    final double Z = std > 0 ? Math.abs((U - (0.5 * n1 * n2)) / std) : 0.;
+    if(std == 0) { return 0; }
```
kno10: Note the std > 0 ternary in the previous version.

elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MWPTest.java
```diff
@@ -139,53 +142,81 @@ private MWPTest() {
...
-    @Override
-    public MWPTest make() {
-      return STATIC;
+
+  protected int getSafeCut(int idx, MWPRanking ranks){
+    double[] ref = ranks.adjusted;
+
+    int i = 0;
+    while(true){
+      // principle of asking for forgiveness instead of permission, exception gets thrown very rarely
```
kno10: I don't like this with respect to readability; and AFAIK setting up an exception handler is not exactly for free (it involves managing an exception table). Also the cost of an exception may be much higher than you expect, as it usually involves a full stack trace (there are some ugly hacks with static exceptions, as seen in ELKIs `ParseUtil`, that may be helpful for common exceptions)
See, e.g., https://shipilev.net/blog/2014/exceptional-performance/

elki-core-math/src/test/java/elki/math/statistics/dependence/mcde/MWPTestTest.java
```diff
@@ -63,4 +62,18 @@ public void testRanking() {
     assertEquals("ranks incorrect", output_duplicates_start_end.adjusted[11], 11.0, 0);
     assertEquals("ranks incorrect", output_duplicates_start_end.adjusted[12], 11.0, 0);
   }
+
+  @Test
+  public void testStatTest(){
+    int len = 10;
+    DoubleArrayAdapter adapter = DoubleArrayAdapter.STATIC;
+    MWPTest mwpTest = MWPTest.STATIC;
```
kno10: As these are used only once, I think they should just be used below without predeclaration.

elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MCDETest.java
```diff
@@ -48,19 +48,25 @@
...
+     */
+    public int len;
+
     /**
      * Constructor.
      *
      * @param idx Array containing index (position) values
      */
-    public RankStruct(int[] idx) {
+    public RankStruct(int[] idx, int len) {
```
kno10: Will the length ever be anything other than idx.length?

elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MWPTest.java
alanmazankiewicz: Done

elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MWPTest.java
alanmazankiewicz: Done

elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MWPTest.java
alanmazankiewicz: Done

elki-core-math/src/test/java/elki/math/statistics/dependence/mcde/MWPTestTest.java
alanmazankiewicz: Done

elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MCDETest.java
alanmazankiewicz: Done

elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MWPTest.java
alanmazankiewicz: I actually don't know why my version did not incorporate this change (as well as the `final double Z = std > 0 ? Math.abs((U - (0.5 * n1 * n2)) / std) : 0.;`) as I fetched and merged the upstream version

elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MWPTest.java
```diff
@@ -50,19 +48,27 @@
...
    * Static Constructor - use this one
    */
   public static final MWPTest STATIC = new MWPTest();
 
+  /**
+   * Share of instances in marginal restriction (reference dimension).
+   * Note that in the original paper alpha = beta and as such there is no
+   * explicit distinction between the parameters.
+   */
+  protected double beta;
```
kno10: Another merge error? beta is not used here.
Maybe you can make a clean rebase on master, and review the differences? Also the empty line before the class name above; and code formatting differences that do not use the ELKI code formatter style.

elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MWPTest.java
```diff
@@ -98,16 +104,16 @@ private MWPTest() {
...
+    int[] idx = MathUtil.sequence(0, len);
+    IntegerArrayQuickSort.sort(idx, (x, y) -> Double.compare(adapter.getDouble(data, x), adapter.getDouble(data, y)));
     MWPRanking I = new MWPRanking(idx);
     long correction = 0;
-    for(int j = 0; j < len;) {
+    for(int j = 0; j < len; ) {
       final double v = adapter.getDouble(data, idx[j]);
       int k = j, t = 1, adjust = 0;
-      while(k < len - 1 && adapter.getDouble(data, idx[k + 1]) == v) {
+      while(k < len - 1 && adapter.getDouble(data, idx[k]) == adapter.getDouble(data, idx[k + 1])) {
```
kno10: Obviously v should still be used here, too, not just stored above.

elki-core-math/src/main/java/elki/math/statistics/dependence/mcde/MWPTest.java
```diff
@@ -178,14 +190,35 @@ public double statisticalTest(int start, int end, boolean[] slice, MWPRanking co
...
-    @Override
-    public MWPTest make() {
-      return STATIC;
+
+  protected int getSafeCut(int idx, MWPRanking ranks) {
+    double[] ref = ranks.adjusted;
+    final int len = ranks.index.length;
+    int i = 0;
+    while(true) {
+      if(idx - i - 1 < 0) {
```
kno10: I'd prefer something like `int i = idx, j = idx;` and then move i backwards and j forwards?
This would make it easier to understand these boundary conditions. They likely could be combined:
```
for (int i=idx, j=idx, last = len - 1; ; i++, j++) {
if (i == 0 || ref[i] != ref[i - 1]) {
  return i;
}
if (j == last || ref[j] != ref[j + 1]) {
  return j;
}
}
```
also, I'd use len = ranks.adjusted.length; and maybe only pass the ranks.adjusted array to the function not the entire object; this makes it easier to see that this only depends on the adjusted ranks.

elki-core-math/src/main/resources/META-INF/elki/elki.math.statistics.tests.mcde.MCDETest
```diff
@@ -0,0 +1 @@
+elki.math.statistics.tests.mcde.MWPTest
```
kno10: Another merge error, this was moved to a different package.

PR #85, opened 2021-04-14
HySortOD source code

BraulioSanchez:
Source code of [HySortOD](https://dl.acm.org/doi/10.1145/3340531.3412033).
The objective is to make a comparison regarding the efficiency of this solution with other state-of-the-art strategies already implemented within the ELKI framework, to be used as an experimental part in the preparation of a survey of outlier detection algorithms.

### DIFF
--- a/elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
+++ b/elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
@@ -65,7 +65,7 @@ import elki.utilities.optionhandling.parameters.IntParameter;
  * 
  * @author Cabral, Eugênio F. (Original Code)
  * @author Braulio V.S. Vinces (ELKIfication)
- * @since 0.1
+ * @since 0.7.5
  *
  * @param <V> Vector type
  */
--- a/elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
+++ b/elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
@@ -29,18 +29,20 @@ import elki.database.Database;
 import elki.database.datastore.DataStoreFactory;
 import elki.database.datastore.DataStoreUtil;
 import elki.database.datastore.WritableDoubleDataStore;
-import elki.database.ids.DBIDIter;
+import elki.database.ids.DBIDArrayIter;
 import elki.database.relation.DoubleRelation;
 import elki.database.relation.MaterializedDoubleRelation;
 import elki.database.relation.Relation;
 import elki.logging.Logging;
 import elki.logging.progress.FiniteProgress;
+import elki.logging.progress.StepProgress;
 import elki.math.DoubleMinMax;
 import elki.outlier.OutlierAlgorithm;
 import elki.result.outlier.BasicOutlierScoreMeta;
 import elki.result.outlier.OutlierResult;
 import elki.result.outlier.OutlierScoreMeta;
 import elki.utilities.Alias;
+import elki.utilities.datastructures.arraylike.IntegerArray;
 import elki.utilities.documentation.Description;
 import elki.utilities.documentation.Reference;
 import elki.utilities.documentation.Title;
@@ -48,6 +50,8 @@ import elki.utilities.optionhandling.OptionID;
 import elki.utilities.optionhandling.Parameterizer;
 import elki.utilities.optionhandling.parameterization.Parameterization;
 import elki.utilities.optionhandling.parameters.IntParameter;
+
+import net.jafama.FastMath;
 
 /**
  * Hypercube-Based Outlier Detection.
@@ -65,7 +69,6 @@ import elki.utilities.optionhandling.parameters.IntParameter;
  * 
  * @author Cabral, Eugênio F. (Original Code)
  * @author Braulio V.S. Vinces (ELKIfication)
- * @since 0.7.5
  *
  * @param <V> Vector type
  */
@@ -84,68 +87,78 @@ public class HySortOD<V extends NumberVector> implements OutlierAlgorithm {
   private static final Logging LOG = Logging.getLogger(HySortOD.class);
 
   /**
-   * Number of bins
+   * Number of bins.
    */
   protected int b;
 
   /**
-   * Threshold to balance the trade-off between the number of hyper-cubes to
-   * scan
-   * during the search and the granularity of mapping
+   * Threshold to balance the trade-off between the number of hypercubes to
+   * scan during the search and the granularity of mapping.
    */
   protected int minSplit;
 
-  // hypercube's length
+  /**
+   * Hypercube's length.
+   */
   private final double l;
 
-  // to match each object with its hyper-cube
-  // it's a little trick to deal with hypercube rearrangement
-  private int[] invertedIndex;
-
-  // strategy for hyper-cubes density search
+  /**
+   * Strategy for hypercubes density search.
+   */
   private DensityStrategy strategy;
 
   /**
-   * Constructor
+   * Constructor with parameters.
    * 
-   * @param k
    * @param b Number of bins
+   * @param minSplit Threshold to balance the tree strategy
    */
   public HySortOD(int b, int minSplit) {
     super();
     this.b = b;
     this.l = 1 / (double) this.b;
     this.minSplit = minSplit;
-    if(this.minSplit > 0) 
+    if(this.minSplit > 0) {
       this.strategy = new TreeStrategy(this.minSplit);
-     else 
+    }
+    else {
       this.strategy = new NaiveStrategy();
+    }
   }
 
   public OutlierResult run(Database db, Relation<V> relation) {
 
     // Output data storage
     WritableDoubleDataStore scores = DataStoreUtil.makeDoubleStorage(relation.getDBIDs(), DataStoreFactory.HINT_DB);
-    FiniteProgress prog = LOG.isVerbose() ? new FiniteProgress("HySortOD scores", relation.size(), LOG) : null;
+    StepProgress stepprog = LOG.isVerbose() ? new StepProgress(3) : null;
+
+    LOG.beginStep(stepprog, 1, "Ordering of hypercubes lexicographically according to their coordinates.");
+    final List<Hypercube> H = getSortedHypercubes(relation);
+
+    LOG.beginStep(stepprog, 2, "Obtaining densities per hypercube.");
+    final int[] W = this.strategy.buildIndex(H).getDensities();
+
     // Track minimum and maximum scores
     DoubleMinMax minmax = new DoubleMinMax();
-
-    int[] W;
-    List<Hypercube> H;
-
     {
-      H = getSortedHypercubes(relation);
-      W = this.strategy.buildIndex(H).getDensities();
+      // compute score by hypercube
+      LOG.beginStep(stepprog, 3, "Computing hypercube scores");
 
-      // Iterate over all objects
-      int objectPosition = 0;
-      for(DBIDIter iter = relation.iterDBIDs(); iter.valid(); iter.advance())
-        scores.putDouble(iter, score(W[this.invertedIndex[objectPosition++]]));
-
-      LOG.incrementProcessed(prog);
+      FiniteProgress prog = LOG.isVerbose() ? new FiniteProgress("HySortOD scores", relation.size(), LOG) : null;
+      DBIDArrayIter iter = (DBIDArrayIter) relation.iterDBIDs();
+      for(int hypercube = 0; hypercube < H.size(); hypercube++) {
+        final IntegerArray instances = H.get(hypercube).getInstances();
+        final double hypercubeScore = score(W[hypercube]);
+        minmax.put(hypercubeScore);
+        for(int instance = 0; instance < instances.size; instance++) {
+          final int index = instances.get(instance);
+          scores.putDouble(iter.seek(index), hypercubeScore);
+          LOG.incrementProcessed(prog);
+        }
+      }
+      LOG.ensureCompleted(prog);
     }
 
-    LOG.ensureCompleted(prog);
     // Wrap the result in the standard containers
     OutlierScoreMeta meta = new BasicOutlierScoreMeta(
         // Actually observed minimum and maximum values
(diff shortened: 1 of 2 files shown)


elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,743 @@
...
+  class Hypercube implements Comparable<Hypercube>, Comparator<Hypercube> {
+
+    int[] coords;
+
+    ArrayList<Integer> instances;
+
+    public Hypercube(V instance, double lenght) {
+      super();
+      this.coords = new int[instance.getDimensionality()];
+      this.instances = new ArrayList<Integer>();
```
kno10, line 214: Please use `IntegerArray` with native integers to avoid the boxing overhead of `Integer.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,743 @@
...
+  protected int b;
+
+  /**
+   * Threshold to balance the trade-off between the number of hyper-cubes to
+   * scan
+   * during the search and the granularity of mapping
+   */
+  protected int minSplit;
+
+  // hypercube's length
```
kno10, line 98: Please use JavaDoc comments here /** */ even for private variables.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,743 @@
...
+   * 
+   * @param k
+   * @param b Number of bins
+   */
+  public HySortOD(int b, int minSplit) {
+    super();
+    this.b = b;
+    this.l = 1 / (double) this.b;
+    this.minSplit = minSplit;
+    if(this.minSplit > 0)
```
kno10, line 119: The ELKI conventions are to even use curly brackets for one-line cases. This can help avoid certain programming errors (e.g., the Apples SSL "goto fail" bug would have been prevented by always using brackets).

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,743 @@
...
+    {
+      H = getSortedHypercubes(relation);
+      W = this.strategy.buildIndex(H).getDensities();
+
+      // Iterate over all objects
+      int objectPosition = 0;
+      for(DBIDIter iter = relation.iterDBIDs(); iter.valid(); iter.advance())
+        scores.putDouble(iter, score(W[this.invertedIndex[objectPosition++]]));
+
+      LOG.incrementProcessed(prog);
```
kno10, line 145: Clearly, this progress logging was meant to be in the loop, or it will be executed only once...

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,743 @@
...
+    OutlierScoreMeta meta = new BasicOutlierScoreMeta(
+        // Actually observed minimum and maximum values
+        minmax.getMin(), minmax.getMax(),
+        // Theoretical minimum and maximum: no variance to infinite variance
+        0, Double.POSITIVE_INFINITY);
+    DoubleRelation rel = new MaterializedDoubleRelation("HySortOD", relation.getDBIDs(), scores);
+    return new OutlierResult(meta, rel);
+  }
+
+  private List<Hypercube> getSortedHypercubes(Relation<V> relation) {
```
kno10, line 159: We also document private methods in ELKI briefly.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,743 @@
...
+        obj.add(objectPosition);
+      }
+
+      objectPosition++;
+    }
+
+    int n = sorted.size();
+    List<Hypercube> H = new ArrayList<Hypercube>(n);
+
+    this.invertedIndex = new int[relation.size()];
```
kno10, line 182: If you used ELKIs DBID data structures above, you would not need this inverted lookup.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,743 @@
...
+
+    public int getDensity() {
+      return instances.size();
+    }
+
+  }
+
+  /**
+   * 
+   * @source https://www.baeldung.com/java-binary-tree
```
kno10, line 300: Including source code copied from some web site is a licensing problem.
But for all I can tell, the BinaryTree may be pretty much a Java TreeSet (which is a red-black tree if I am not mistaken).

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,743 @@
...
+      return root.left == null ? root.value : findSmallestValue(root.left);
+    }
+
+    @Override
+    public Iterator<T> iterator() {
+      return new TreeIterator(root);
+    }
+
+    /**
+     * @source https://stackoverflow.com/a/12851421
```
kno10, line 417: Again a licensing problem. This code is supposedly CC BY-SA licensed.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,743 @@
...
+ * <p>
+ * Reference:
+ * <p>
+ * Cabral, Eugênio F., Robson L.F. Cordeiro<br>
+ * Fast and Scalable Outlier Detection with Sorted Hypercubes<br>
+ * Proc. 29th ACM Int. Conf. on Information & Knowledge Management (CIKM'20)
+ * 
+ * @author Cabral, Eugênio F. (Original Code)
+ * @author Braulio V.S. Vinces (ELKIfication)
+ * @since 0.7.5
```
kno10, line 68: The `@since` annotations at which release(!) a class was first available. We insert them automatically at a release. Hence on new contributions, this annotation should not be present.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
kno10: But I guess most work is done in the hypercube methods above, a progress bar would be more helpful there.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+
+    /**
+     * Minimum number of rows to allow sub-mapping.
+     */
+    final int minSplit;
+
+    /**
+     * Maximum number of dimensions to map.
+     */
+    int numMappedDimensions;
```
kno10, line 465: This variable is unused.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+  public TypeInformation[] getInputTypeRestriction() {
+    return TypeUtil.array(TypeUtil.NUMBER_VECTOR_FIELD);
+  }
+
+  /**
+   * Bounded regions of the space where at least one instance exists.
+   * 
+   * @author Cabral, Eugênio F.
+   */
+  class Hypercube implements Comparable<Hypercube> {
```
kno10, line 223: The class can be made `static`, which makes the objects smaller (no pointer to parent).

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+    protected int b = 5;
+
+    protected int minSplit = 100;
+
+    @Override
+    public void configure(Parameterization config) {
+      new IntParameter(B_ID, 5) //
+          .grab(config, x -> b = x);
+      new IntParameter(MIN_SPLIT_ID, 100) //
+          .grab(config, x -> minSplit = x);
```
kno10, line 685: min splits needs to be >= 0, right?

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+     */
+    public static final OptionID MIN_SPLIT_ID = new OptionID("hysortod.minsplit", "Predefined threshold to use.");
+
+    protected int b = 5;
+
+    protected int minSplit = 100;
+
+    @Override
+    public void configure(Parameterization config) {
+      new IntParameter(B_ID, 5) //
```
kno10, line 682: b needs to be at least 1, right? Or at least 2 to be meaningful?

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+        int midVal = this.H.get(i).getCoordAt(col);
+        int rgtVal = this.H.get(i).getCoordAt(col) + 1;
+
+        Node lftNode = parent.childs.get(lftVal);
+        Node midNode = parent.childs.get(midVal);
+        Node rgtNode = parent.childs.get(rgtVal);
+
+        int nextCol = Math.min(col + 1, this.H.get(i).getNumDimensions() - 1);
+
+        if(Objects.nonNull(lftNode)) {
```
kno10, line 594: Object.nonNull is appropriate as a functional predicate in a filter() call.
Here, `lftNode != null` is more appropriate.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+
+    /**
+     * Hypercube coordinates.
+     */
+    final int[] coords;
+
+    /**
+     * Holds a set of instances within the hypercube.
+     */
+    final IntegerArray instances;
```
kno10, line 233: Using a ModifiableDBIDArray and storing the ids directly is better here.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+   * 
+   * @param relation Data to process
+   * @return Hypercubes sorted
+   */
+  private List<Hypercube> getSortedHypercubes(Relation<V> relation) {
+    TreeSet<Hypercube> sorted = new TreeSet<HySortOD<V>.Hypercube>();
+
+    // Iterate over all objects
+    for(DBIDArrayIter iter = (DBIDArrayIter) relation.iterDBIDs(); iter.valid(); iter.advance()) {
+      Hypercube h = new Hypercube(relation.get(iter).toArray(), this.l);
```
kno10, line 183: This creates a copy of each vector, then a hypercube object for it, a fairly expensive operation.
Likely the algorithm could be made much faster if the lookup would not require object creation.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+      return density;
+    }
+
+    /**
+     * Tree node.
+     * 
+     * @author Cabral, Eugênio F.
+     *
+     */
+    private class Node {
```
kno10, line 614: Can trivially be made static, to avoid storing a pointer to the parent in every node.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+
+      /**
+       * Index information.
+       */
+      public final int value, begin, end;
+
+      /**
+       * Childs of the node.
+       */
+      public final HashMap<Integer, Node> childs;
```
kno10, line 624: Using a Int2ObjectOpenHashMap<Node> avoids boxing overheads.
But it may be worth instantiating the map only when used.

kno10: Also childs -> children

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+     */
+    final IntegerArray instances;
+
+    /**
+     * Hypercube constructor.
+     *
+     * @param values Vector representation of an instance
+     * @param lenght Hypercube's length
+     */
+    public Hypercube(double[] values, double lenght) {
```
kno10, line 241: Allowing a `NumberVector values` instead of the `double[]` avoids the copy trivially.

kno10: Then I could even remove the <V> generic from the class altogether.

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+      this.coords = new int[values.length];
+      this.instances = new IntegerArray();
+
+      for(int d = 0; d < values.length; d++) {
+        this.coords[d] = (int) FastMath.floor(values[d] / lenght);
+      }
+    }
+
+    @Override
+    public int compareTo(Hypercube o) {
```
kno10, line 252: Can be simplified to `return Arrays.compare(coords, o.coords)`

elki-outlier/src/main/java/elki/outlier/density/HySortOD.java
```diff
@@ -0,0 +1,694 @@
...
+        int d = this.coords[i] - o.coords[i];
+        if(d != 0) {
+          return d;
+        }
+      }
+      return 0;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
kno10, line 263: Can be simplified a lot by using `Arrays.equals(coords, o.coords)`

kno10:
Merged with changes in b0d9607554d1d7a4d37fd8a0eb1fbef372eaa052

kno10:
Merged with some optimizations. There is some potential for further acceleration by avoiding creating as many `HyperCube` objects, and instead implementing a more direct index tree.

PR #89, opened 2021-05-18
Hysortod

BraulioSanchez:


kno10:
Why should this be more efficient?

BraulioSanchez:
Yes, we have a benchmark considering all the data sets used in the article. The efficiency of this algorithm is sensitive to the number of dimensions. We use a data set of 25 dimensions and 25k instances, we compare the execution time of this implementation (14 sec) with the original one (4 sec). The bottleneck is in the getDensities method of the TreeStrategy class.

kno10:
Sorry that I have not been more responsive. Too many duties.
Your patch undoes two changes: (i) lazy initialization of the children map, (ii) null checks location in the density function.
Did you benchmark both separately? It's always a bit tricky to predict Hotspot performance. Often the shorter code yields shorter bytecode, which may end up being optimized better; that is why I did this one-liner for the recursive calls.

Can you try 00824854dc7a4daa0489ce3513434a6cd202a57c with your benchmark suite? Thank you.

PR #92, opened 2021-06-27
source code of Isolation Forest algorithm

BraulioSanchez:


elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+  public IsolationForest(int numTrees, int subSamplingSize, RandomFactory rnd) {
+    this.numTrees = numTrees;
+    this.subsampleSize = subSamplingSize;
+    this.rnd = rnd;
+  }
+
+  public OutlierResult run(Database db, Relation<? extends NumberVector> relation) {
+    // Output data storage
+    WritableDoubleDataStore scores = DataStoreUtil.makeDoubleStorage(relation.getDBIDs(), DataStoreFactory.HINT_DB);
+    FiniteProgress prog = LOG.isVerbose() ? new FiniteProgress("IsolationForest progress", relation.size(), LOG) : null;
```
kno10, line 131: The `prog` progress should run only until numTrees, not `relation.size()`, and you'll likely want a second progress for scoring.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+        LOG.incrementProcessed(prog);
+      }
+      LOG.ensureCompleted(prog);
+
+      // Iterate over all objects
+      for(DBIDIter iter = relation.iterDBIDs(); iter.valid(); iter.advance()) {
+        final double score = distributionForInstance(relation.get(iter));
+        scores.putDouble(iter, score);
+        minmax.put(score);
+        LOG.incrementProcessed(prog);
```
kno10, line 161: Do not re-use a progress that you already "completed" above. This is a second phase, and one step in this progress is much less effort than one tree construction above.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+
+  /**
+   * Returns the average path length of an unsuccessful search. Returns 0 if
+   * argument is less than or equal to 1.
+   */
+  protected static double c(double n) {
+    if(n <= 1.0) {
+      return 0;
+    }
+    return 2 * (FastMath.log(n - 1) + 0.5772156649) - (2 * (n - 1) / n);
```
kno10, line 184: Undocumented "magic" constants such as `0.5772156649` are considered a bad style. Please replace this with a named constant from `MathUtil` (I don't think we have the Euler constant there yet).

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+      // Set size of node
+      size = data.size();
+
+      // Stop splitting if necessary
+      if((size <= 1) || (height == maxHeight)) {
+        return;
+      }
+
+      // Compute mins and maxs and eligible attributes
+      final int dim = RelationUtil.dimensionality(relation);
```
kno10, line 246: While this is O(1) and still fairly cheap, it involves a few type checks. It likely is best to pass along the dimensionality as additional parameter.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+      double[][] minmax;
+      {
+        double[] mins = new double[dim], maxs = new double[dim];
+        for(int i = 0; i < dim; i++) {
+          mins[i] = Double.MAX_VALUE;
```
kno10, line 251: Use `Arrays.fill(mins, Double.MAX_VALUE)` instead.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+          maxs[i] = -Double.MAX_VALUE;
+        }
+        for(DBIDIter iditer = data.iter(); iditer.valid(); iditer.advance()) {
+          final NumberVector o = relation.get(iditer);
+          for(int d = 0; d < dim; d++) {
+            final double v = o.doubleValue(d);
+            mins[d] = Math.min(v, mins[d]);
```
kno10, line 258: If `v` is NaN, this will yield a `NaN`.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+            maxs[d] = Math.max(v, maxs[d]);
+          }
+        }
+        minmax = new double[][] { mins, maxs };
+      }
+
+      ArrayList<Integer> al = new ArrayList<Integer>();
```
kno10, line 265: Never use `Collection<Integer>`, it is *very* expensive due to boxing.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+        if(minmax[0][j] < minmax[1][j]) {
+          al.add(j);
+        }
+      }
+
+      // Check whether any eligible attributes have been found
+      if(al.size() == 0) {
+        return;
+      }
+      else {
```
kno10, line 276: No need for the `else`, because of the `return` above.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+        // Randomly pick an attribute and split point
+        a = al.get(r.nextInt(al.size()));
+        q = (r.nextDouble() * (minmax[1][a] - minmax[0][a])) + minmax[0][a];
+
+        // Create sub trees
+        successors = new ArrayList<>(2);
+        for(int i = 0; i < 2; i++) {
+          ArrayModifiableDBIDs tmpData = DBIDUtil.newArray();
```
kno10, line 284: Allocating the array with `data.size()` is likely a bit faster, and the temporary additional memory usage should be okay.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+        }
+        for(DBIDIter iditer = data.iter(); iditer.valid(); iditer.advance()) {
+          final NumberVector o = relation.get(iditer);
+          for(int d = 0; d < dim; d++) {
+            final double v = o.doubleValue(d);
+            mins[d] = Math.min(v, mins[d]);
+            maxs[d] = Math.max(v, maxs[d]);
+          }
+        }
+        minmax = new double[][] { mins, maxs };
```
kno10, line 262: Why not keep mins, and maxs? Nothing is gained by nesting them.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+        return;
+      }
+      else {
+        // Randomly pick an attribute and split point
+        a = al.get(r.nextInt(al.size()));
+        q = (r.nextDouble() * (minmax[1][a] - minmax[0][a])) + minmax[0][a];
+
+        // Create sub trees
+        successors = new ArrayList<>(2);
+        for(int i = 0; i < 2; i++) {
```
kno10, line 283: Remove this loop, and construct the two children in one pass.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+          ArrayModifiableDBIDs tmpData = DBIDUtil.newArray();
+          for(DBIDIter iditer = data.iter(); iditer.valid(); iditer.advance()) {
+            NumberVector vector = relation.get(iditer);
+            if((i == 0) && (vector.doubleValue(a) < q)) {
+              tmpData.add(iditer);
+            }
+            if((i == 1) && (vector.doubleValue(a) >= q)) {
```
kno10, line 290: Use two tmpData, and remove the outer loop.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+        a = al.get(r.nextInt(al.size()));
+        q = (r.nextDouble() * (minmax[1][a] - minmax[0][a])) + minmax[0][a];
+
+        // Create sub trees
+        successors = new ArrayList<>(2);
+        for(int i = 0; i < 2; i++) {
+          ArrayModifiableDBIDs tmpData = DBIDUtil.newArray();
+          for(DBIDIter iditer = data.iter(); iditer.valid(); iditer.advance()) {
+            NumberVector vector = relation.get(iditer);
+            if((i == 0) && (vector.doubleValue(a) < q)) {
```
kno10, line 287: Use <=q first, >q below, to avoid the first child to ever be empty.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+          }
+          successors.add(new Tree(tmpData, r, height + 1, maxHeight, relation));
+        }
+      }
+    }
+
+    /**
+     * Returns path length according to algorithm.
+     */
+    protected double pathLength(NumberVector vector) {
```
kno10, line 302: An iterative version of this is likely more performant than a recursive version (which allocates stack frames).
Just use `while(successor != null) { depth++; successor = successors.get(...) }`

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+  public TypeInformation[] getInputTypeRestriction() {
+    return TypeUtil.array(TypeUtil.NUMBER_VECTOR_FIELD);
+  }
+
+  /**
+   * Parameterization class
+   * 
+   * @hidden
+   * 
+   * @param <V> Vector type
```
kno10, line 325: There is no `<V>` here, hence, remove this line.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+     * Parameter for the sub sample size
+     */
+    public static final OptionID SUBSAMPLE_SIZE_ID = new OptionID("iforest.subsample", "Sub sample size.");
+
+    /**
+     * Parameter to specify the seed to initialize Random.
+     */
+    public static final OptionID SEED_ID = new OptionID("iforest.seed", "The seed to use for initializing Random.");
+
+    protected int numTrees = 100;
```
kno10, line 343: Please add JavaDoc documentation to these fields.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+
+    protected int subsampleSize = 256;
+
+    protected RandomFactory rnd;
+
+    @Override
+    public void configure(Parameterization config) {
+      new IntParameter(NUM_TREES_ID, 100) //
+          .addConstraint(CommonConstraints.GREATER_THAN_ONE_INT) //
```
kno10, line 352: Why >1? I'd assume `GREATER_EQUAL_ONE_INT`.

elki-outlier/src/main/java/elki/outlier/isolation/IsolationForest.java
```diff
@@ -0,0 +1,366 @@
...
+          .grab(config, x -> numTrees = x);
+      new IntParameter(SUBSAMPLE_SIZE_ID, 256) //
+          .addConstraint(CommonConstraints.GREATER_THAN_ONE_INT) //
```
kno10, line 355: Here 2 is probably still a rather meaningless minimum size, where we would expect all scores to be the same, right?

kno10:
I merged an own implementation of isolation forest (similar quality in the test, but more efficient implementation).

PR #93, opened 2021-07-18
Source code of GLOSH score

BraulioSanchez:
In the `GLOSH` class there is a `TODO`, could you please tell me how to fix it?


BraulioSanchez:
I need to fix this implementation.

PR #94, opened 2021-07-20
Source code of KMeansMinusMinusOutierDetection

BraulioSanchez:
Source code of `KMeansMinusMinusOutierDetection` for identifying noise points as outliers, noise flag removed from `KMeansMinusMinus` to keep the original publication proposal.


kno10:
I'd rather keep the noise flag option, the default is still the original publication.

kno10:
As for the current code, I don't know if we shouldn't solve this differently:
Right now, the code produces a binary outlier label, which is effectively 1 exactly if objects are in a noise cluster. We could write a "NoiseAsOutliers" class that would work both with k-means-- as well as DBSCAN and perform this transformation.
But it would likely be more in line with k-means-- – which ranks objects by the distance to the nearest cluster centers – to produce a score based on the distance to the cluster center, i.e. use KMeansOutlierDetection with k-means-- and assign "noise" points to the nearest cluster (i.e., without kmeansmm.noisecluster). Which also allows comparing regular k-means and k-means-- consistently.

PR #95, opened 2021-07-29
Added TBA-DBSCAN

flexkra:
Added source code of TBA-DBSCAN based on the master thesis "Beschleunigung von DBSCAN mithilfe der Dreiecksungleichung" by Felix Krause


kno10:
We (mostly) include code that is either published or at least on its way to be published; so for now its not going to be merged as is.
An implementation closer to published TI-DBSCAN would be a candidate for merging, though.

PR #97, opened 2021-08-21
Source code of GLOSH score

BraulioSanchez:


kno10:
Hello,
This pull likely needs to be recreated, because much of the hierarchical clustering (including HDBSCAN) was rewritten to use a merge history data structure instead of the SLINK-type pointer hierarchy previously used. That rewrite is finished, and seems to work fine. But this pull request now no longer can be merged, can you maybe update it?
Thank you.

kno10:
Merged an adapted version

PR #101, opened 2022-06-03
Fix StackOverflow ex in AlphaShape

SIDSSIDS:
Replaced usage of recursion with a stack in the alpha shape computation.
Recursion led to StackOverflow exception on big datasets (4646 points in my case)

### DIFF
--- a/elki-core-math/src/main/java/elki/math/geometry/AlphaShape.java
+++ b/elki-core-math/src/main/java/elki/math/geometry/AlphaShape.java
@@ -20,7 +20,9 @@
  */
 package elki.math.geometry;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Deque;
 import java.util.List;
 
 import elki.data.spatial.Polygon;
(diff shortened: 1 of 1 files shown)


kno10:
Do you have an interesting test case for benchmarking?

SIDSSIDS:
> I'd avoid the nested data structure, but use an IntegerArray

By _nested data structure_ you mean constructions like `new int[]{i, tri.ca, tri.c, tri.a}`?
I'm not sure, that this kind of array initialization is slower than the following:
```
int[] a = new int[4];
a[0] = i;
a[1] = tri.ca;
a[2] = tri.c;
a[3] = tri.a;
```
If so, can you give me a hint why?

> Do you have an interesting test case for benchmarking?

I have only the test case, which has led to the problem.
[alpha_points.csv](https://github.com/elki-project/elki/files/9009518/alpha_points.csv)
You can reproduce the error with the default value of alpha
```
java -Xss512k -jar elki-bundle-0.7.6-SNAPSHOT.jar KDDCLIApplication -dbc.in alpha_points.csv -algorithm algorithm.NullAlgorithm -resulthandler visualizer
```
and pick `DoubleVector` -> `Scatterplot` -> `Alpha Shape`

kno10:
Using a single `IntegerArray` array as stack (a wrapped `int[]`), and the four values stored at offsets `(i<<2) + j`.
See 6465675 for the resulting code.
This is based on your contribution, although I used a different data structure that uses less memory. Its not pretty code, because the four array adds are unrolled. I could have used an "add4" method, which likely would be inlined and hence not make much of a difference.
Thank you for the test case for reproducing the error (as this allowed me to check it is resolved).

PR #103, opened 2022-07-03
Outlierness score considering a single cluster or noise

BraulioSanchez:
This commit attempts to correct the attribution of anomaly scores in cases where there is only one group or noise.


kno10:
Solved this slightly differently, because it not only affects outlier detection - also for regular Silhouette evaluation, the case of a single cluster should be handled. But thank you for the report & proposed fix!

